PR #1, opened 2019-10-12
image readme & subfolders

neablis-7:
Added a readme describing the process to create images for cards, as well as subfolders for each category of cards. The readme for each folder describes the cards that should be in that folder.


neablis-7:
I fixed the folder names to have underscores instead of spaces.

PR #2, opened 2019-10-13
Extracted local paths to settings.xml

HalfTheDeck:
With these changes pom.xml does not have to be edited if the paths are not the same as in the example, and thus will not appear as a local change or even be committed by accident.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,13 +21,30 @@ Discussion in the #jorbs-spire-mod-char channel in the [Jorbs Discord](https://d
 
 ## How to build from source
 
-1. Create an environment variable named `STEAM_PATH` set to `C:/Program Files/Steam/steamapps` (or wherever you have Steam installed) 
 1. Install the Oracle Java 8 JDK (not Java 9+)
-1. Install IntelliJ IDEA (the free Community edition is fine)
-1. Clone the repository
-1. Open the project in IntelliJ IDEA:
+2. Install IntelliJ IDEA (the free Community edition is fine)
+3. Clone the repository
+4. Open the project in IntelliJ IDEA:
     1. Choose "Import Project"
     1. Choose the repository folder
     1. Select "Maven"
     1. Press next a few times, all other settings can be left at defaults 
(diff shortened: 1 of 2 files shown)


dbjorge:
I don't really understand the advantage of using a file rather than the environment variable it was already using, but I'm willing to trust that you had some reason for wanting to switch it.

HalfTheDeck:
Thanks for trusting.

The environment variable only covers Steam installation paths. On Macs, for instance, the path to the STS jar is also different. I was looking for a way to extract those local properties to another file, which is not under version control. I would have preferred a .properties file in the project folder, which I would have added to gitignore. But that is apparently not the maven way. Then I found the settings.xml solution.

I'm an experienced software developer/architect and have been and am working on many Java projects (in private company repositories, that's why you can't see them). I want to help developing this mod. Can you please add me to the team? I don't want to create a fork every time I work on the mod. I'd rather create branches, submit pull requests and delete the branches after they have been merged.

HalfTheDeck:
Oh, now I'm the third contributor. Didn't update until now. So I guess the contribution process involves forking, submitting pull requests and merging the master back into the fork after the pull requests have been merged. I'm OK with that. So if you need my help, hit me up.

dbjorge:
@HalfTheDeck Awesome, thanks! I've added you to the authors list, too. The workflow I recommend using with a GitHub fork+PR based project like this is:

```sh
# Prerequisite: On GitHub, fork this repository to your own user account

# first-time setup
git clone --origin upstream https://github.com/dbjorge/jorbs-spire-mod.git
cd jorbs-spire-mod
git remote add my_fork https://github/your_username/jorbs-spire-mod.git

# When starting a new feature, start a new branch based on the current upstream master version
# (in this workflow, you don't use your fork's master at all)
git checkout master
git pull
git checkout -b your_new_feature

# make your changes, git add/git commit as usual

# When feature is ready to merge...
git push -u my_fork HEAD
# On GitHub, make a pull request from your_username:your_new_feature to dbjorge:master
```

PR #5, opened 2019-10-15
Enqueue actions

HalfTheDeck:
This is what I had in mind. Usage examples would be (don't know/care if the enum values actually exist or the actions actually make sense - could also be shorter even with already existing methods):

new EnqueueActions(monster)
.setDefaultTargetDamage(10, DamageType.POISON, AttackEffect.SPLASH)
.calculateAndDealDamageToTarget();

with Clarity mod (shows broken chaining due to subclassing):

EnqueueWandererActions enqueueActions = new EnqueueWandererActions(monster)
.applyClarityAddend(0);
enqueueActions.setDefaultTargetDamage(10, DamageType.FIRE, AttackEffect.BURN)
.setDefaultSourceDamage(1, DamageType.FIRE, AttackEffect.BURN)
.calculateAndDealDamageToSource()
.calculateAndDealDamageToTarget();

enqueueActions.applyClarityMultiplier(2)
.resetSourceDamage()
.setDefaultTargetDamage(3, DamageType.LIGHTNING, AttackEffect.FLASH)
.calculateAndDealDamageToTarget()
...

If this is OK, I will add more methods (AoE, healing etc.).


dbjorge:
Hi @HalfTheDeck,

I appreciate the suggestion, but I'm not sure I'm understanding correctly what the intended usage would look like in practice. I tried updating an existing card implementation that seemed like a good fit for what you've implemented so far (`Firebolt: Deal 8 damage + 2(3) damage for each Clarity`), but it didn't seem like it made the implementation more readable:

* the `use` implementation ended up being the same length
* it was harder to understand at a glance what the order of operations was; if I were reading the implementation for the first time, I'd have to dig into the builder implementation to understand the order of operations for the "clarity multiplier" vs the base damage, whereas in the current implementation, it's clear without jumping between classes
* it didn't help with the `calculateCardDamage` or `applyPowers` cases; these are important not just for calculating the damage action, but also because they interact with how the game dynamically updates card descriptions as it calculates the impact of weaken/vulnerable/etc

I suspect I'm misunderstanding how you're intending to apply the abstraction. Would it be possible for you to show a more complete example of how you would expect this to apply to a real card (eg, `Firebolt`), so I can understand the problem this is trying to solve?

---

As an aside, you can avoid the broken chaining issue by using the curiously recurring template pattern:

```java
public abstract class BaseEnqueueActions<SelfT extends BaseEnqueueActions<SelfT>> {
    // ... current impl of EnqueueActions, except builder methods return type SelfT by calling "return (SelfT)this;"
}
public class EnqueueActions extends BaseEnqueueActions<EnqueueActions> { /* forwarding ctors */ };
public class WandererEnqueueActions extends BaseEnqueueActions<WandererEnqueueActions> {
    // ... current impl of WandererEnqueueActions ...
}

```

...but I'm more interested in understanding what the desired usage would look like first, before worrying about the implementation.

dbjorge:
Your PR definitely encouraged me to think about whether there were any simplifications that could be made, refactoring-wise, so I've gone ahead and added several improvements in that area. In particular, the relevant bits of `Firebolt`'s implementation are now reduced to:

```java
    @Override
        return this.magicNumber * MemoryPowerUtils.countClarities(AbstractDungeon.player);
    }

    @Override
        enqueueAction(new DamageAction(m, new DamageInfo(p, damage), AttackEffect.FIRE));
    }
```

I think this is pretty short and understandable as-is, but I'm definitely happy to look over any suggestions you have to improve it further!

HalfTheDeck:
Hey @dbjorge,

I'm glad I inspired you. I was also initially considering an inheritance approach, where the methods in `EnqueueActions` are provided by `AbstractDefaultCard` or `AbstractDynamicCard` or new subclasses. But in this case I preferred the decoupled implementation, because cards aren't the only sources of actions.

The examples, where the immediate benefits of EnqueueActions become more obvious, are indeed more complex, than what I have written so far. It originally wasn't intended for simple implementations of `use`, where it would just abstract away `addToTop` and its parameters. Currently there are three cards, where the implementation of `use` could be replaced: `Hurt`, `MagicMissles` (btw: Is there a reason why the second 'i' is missing in missiles?) and `PoisonSpray`, where `Hurt` is the only one, that is currently covered (have to add a `>0` check, though).

The implementations of `update`, particularly in `ChainLightningAction` and `SnapAction`, are more natural/suitable targets.

I have to admit, that 5 classes are not very many to justify this construct, which will probably be relatively big once it's finished. But there are more cards and at least 2 more characters to come.

Initially, it was intended for the execution part of a card's effect, only. For updating the card description, this would have to be more integrated into the `AbstractDynamicCard` class with a dry-run case and the actual execution case. I can see into that.

However, there is one big reason to use the abstraction where possible: If Megacrit were to change their method signatures, all cards and actions would have to be changed. I don't intend to cover every game action, but if at least all `DamageAction`s were abstracted away, this problem would be reduced to an acceptable amount of necessary edits.

Fun fact: I have used the pattern, you mentioned, 12 years ago without knowing the pattern. It was a fun usage of generics at that time. I was even thinking about implementing it here after I had written the classes, but I decided not to, because it was too much effort for too little benefit in the current stage of the implementation.

HalfTheDeck:
I think I found a reasonable way to integrate `EnqueueActions` into the new base card class. It is completely optional (that could be changed easily, though). If a programmer doesn't want to use it, they just implement the subclass for their card like before. But they miss out on the benefits of declaring their actions in one method, marking the calculated damage if necessary, and everything else is taken care of automatically.

I also changed the method names for applying clarity values, because I wasn't really happy with them. Still didn't implement the pattern which avoids breaking chains, because it does nothing right now. Will do it later, if you accept my suggestion.

dbjorge:
> (btw: Is there a reason why the second 'i' is missing in missiles?)

Nope, just a typo. Fixed now, thanks for noting it!

> However, there is one big reason to use the abstraction where possible: If Megacrit were to change their method signatures, all cards and actions would have to be changed. I don't intend to cover every game action, but if at least all DamageActions were abstracted away, this problem would be reduced to an acceptable amount of necessary edits.

I disagree that this would be a good tradeoff to justify an additional layer of abstraction, mostly for two reasons:
* Megacrit is aware that they have a large modding community, and I think the risk of them making a breaking change like this to a commonly used action is relatively low. For example, they have done a reasonable job of avoiding many breaking changes like this between the current stable and current beta branches, one of the largest content patches they've released.
* If they *were* to making a breaking change to `DamageAction`, it would *already* be an acceptable amount of edits to fix without this layer of abstraction (probably it would be a ctrl+shift+f fix; at worst, it would be "ctrl+shift+F 'new DamageAction' -> 'DamageActionAdapter.FromLegacyArgs', create an adapter method that deals with the adaptations, refactor -> inline method").

> I have to admit, that 5 classes are not very many to justify this construct, which will probably be relatively big once it's finished. But there are more cards and at least 2 more characters to come.

I really do appreciate anyone making suggestions or sending ideas, but I just don't think this abstraction is justifiable, even making conservative assumptions about the number of cards and characters to come. Those relevant implementation of those 5 cases combined are 61 lines; even this incomplete form of `EnqueueActions`, which you've said covers **part** of **one** of those 5 cases, is already about 6 times as much code and adds multiple new layers of abstraction on top of implementations that just aren't complicated enough or common enough to warrant abstracting at this point.

I definitely think it's possible that other sets of cards/actions in the future might end up sharing enough functionality that it would be worth making abstractions for common code, but I don't think it's a good idea to try to guess at the right abstractions preemptively; I would much rather see abstractions like this driven by how they would be used in practice (ie, write the code that would use the helpers in practice *first*, and reach satisfication that it actually solves a common problem in a common way (maybe even send a PR/message at that point, to discuss the suggested design), and *only then* deal with implementation).

I hope this doesn't come across as too negative; again, I appreciate the idea, but I think at this point this abstraction would add more overhead than it reduces.

PR #11, opened 2019-10-17
Action configuration

HalfTheDeck:
I've been working on this a little more before and after you closed the other PR. Now the existing code is completely independent of it.

My goal was to be able to declaratively define a card. Ideally, in order to implement a new card, you just extend one of the new abstract base classes - `ConfiguredJorbsModCard` for cards where the actions' base values depend on the target monster and `TargetIndependentConfiguredJorbsModCard` where the actions' base values are always the same regardless of the card's target - and implement the abstract methods only. Base damage calculation and action enqueuing is being taken care of by the base class using the `ActionConfiguration`, which is created and configured in the implementation of the abstract methods.

Creating new cards like this saves a lot of time. Programmers don't have to look up all the available `AbstractGameAction` subclasses. A lot of cards, I'd say the vast majority, will be covered by this and can be written (and maintained) in no time.

The cost for this is the time to write that little framework. But now is the perfect time for doing so. The cards are probably still subject to changes. At least two new characters are yet to come. When it comes to implementing cards, I really wanna be able to use that.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <groupId>stsjorbsmod</groupId>
     <artifactId>JorbsMod</artifactId>
     <name>Jorbs Mod</name>
-    <version>0.9.0</version>
+    <version>0.10.0</version>
     <description>A mod based on Jorbs' spreadsheet.</description>
 
     <properties>
--- a/src/main/java/stsjorbsmod/actions/CounterspellAction.java
+++ b/src/main/java/stsjorbsmod/actions/CounterspellAction.java
@@ -12,7 +12,7 @@ import com.megacrit.cardcrawl.vfx.ThoughtBubble;
 import stsjorbsmod.JorbsMod;
 
 public class CounterspellAction extends AbstractGameAction {
-    private static final String UI_ID = JorbsMod.makeID(CardsToTopOfDeckAction.class.getSimpleName());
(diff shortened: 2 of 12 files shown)


dbjorge:
I appreciate that you are spending time to try to help, but to be frank, I am becoming frustrated because I feel that I am repeating myself with each new batch of changes. I believe that every point I made as I closed the last PR still applies to this one; like the last PR, I think this is a solution in search of a problem.

HalfTheDeck:
@dbjorge 

I strongly disagree. You're basically saying, that all I've been doing the last 10 odd years is being in search of a problem. No, this is preparation at the beginning a software project. The result in all projects, I have been working in, was a well-structured, stable and easily maintainable product.

Well, you can also do it like a hobby coder, with a lot of copy & pasting. Where bug fixing takes ages and maintenance becomes more and more problematic until it won't be possible at all. If you choose this way, good luck. You'll need it. I can't work like that.

dbjorge:
I have not said that I am against abstracting common code; I have that I would like to see such abstractions driven by actual usage and that I haven't thought the *proposed* abstractions would save more complexity than they introduce, and you have not demonstrated otherwise with concrete examples despite repeated requests for them.

Giving "YAGNI" feedback on a few specific PRs is not the same as besparging your entire career's work. Neither the assumption of ill intent nor the passive-aggressive tone are acceptable behavior here; I'm going to have to ask you to either conduct yourself more professionally or leave.

PR #13, opened 2019-10-18
Charity strength refactor

### LINKED ISSUE #8: Charity reduces damage of all attacks
I think this is related to when it applies damage changes. Attacks end up dealing 0 damage if the player has less than 100 gold.

mdiehr:
Now buffs/unbuffs strength appropriately.
Fixes #8


dbjorge:
I added a separate fix for #8 (https://github.com/dbjorge/jorbs-spire-mod/commit/f03f67608dd4ea9fa5f4230fa2b5754139291893), but I think this PR is still interesting. I debated about whether to use an actual `StrengthPower` or not for the Charity effect (the original implementation actually did this, though it was a less complete version that didn't update with mid-fight gold changes), but I switched to the `atDamageGive` form because I thought it would be undesirable for Artifact to prevent the unbuff-on-forget action. With #8 being fixed separately, do you have any other reasons in mind to prefer one strategy over the other?

PR #18, opened 2019-10-19
Temp

HalfTheDeck:
I've been thinking long and hard if I should open yet another PR, because the risk of this being for nothing again is quite high. But I came to the conclusion that it's worth it.

I provided an example: `ConfiguredWeightOfMemory`. I placed it in the `actions.configuration` package on purpose, even though it's a card. Real card implementations (as opposed to examples), of course, still go to the `cards` package.

I had to add the Remember Memory card multiplier to `WandererActionConfiguration`.

In your implementation of `use()`, you apply the value of `damage`. I suppose this is the value which has been calculated in `calculateCardDamage()` or `applyPowers()`. If you just look at the class, you could think `damage` is the same value as `baseDamage`, because there is no other assignment. In the configured version of that card, there can be no confusion about the calculated and actually dealt damage.

Moving the static methods for calculating the number of Remember Memory cards to the configuration class also simplified the card class.

When I was implementing the example, the same thing happened, that always happens, when I design the framework upfront. When actually using the framework, I very rarely have to change its API if something is wrong. The majority of the time, I just fix the implementation of the framework and it works. To me, the card configuration framework is conceptually sound and thus the risk of having to change its API, breaking client code, tends to 0. Of course, it will be extended, but that doesn't break client code. I also think that I found the final names for the `add` and `multiply` methods.


dbjorge:
> Moving the static methods for calculating the number of Remember Memory cards to the configuration class also simplified the card class.

I think deciding whether the card-specific "count remember cards" method should be moved out of the card that it's specific to is orthogonal to the design (you could just as easily imagine moving it to `MemoryUtils.countRememberMemoryCards()`), so I think the meaningfully different chunks of the two cards implementations are:

```java
@Override
    return this.magicNumber * MemoryUtils.countRememberMemoryCards();
}

@Override
    enqueueAction(new DamageAction(m, new DamageInfo(p, damage), AttackEffect.BLUNT_HEAVY));
}
```

vs 

```java
@Override
protected ActionConfiguration createActionConfiguration() {
    return new WandererActionConfiguration()
        .multiplyWithNumberOfRememberMemoryCards(this.magicNumber)
        .calculateAndDealDamageToTarget(baseDamage, null, AttackEffect.BLUNT_HEAVY);
}
```

---

> In your implementation of use(), you apply the value of damage. I suppose this is the value which has been calculated in calculateCardDamage() or applyPowers(). If you just look at the class, you could think damage is the same value as baseDamage, because there is no other assignment. In the configured version of that card, there can be no confusion about the calculated and actually dealt damage.

I agree that the base game's `AbstractCard` implementation forces the user to understand the differences and temporal coupling between `damage` and `baseDamage` in order to use the two correctly, but I don't think your design solves that problem; the user in the second example **has still had to choose whether to use damage or baseDamage**, so they *still* need to understand the difference between the two of them, they just *additionally* now need to understand how they apply to this new extra framework (eg, the questions I had when looking at this code included "does the new framework include vulnerable/weak in its calculations?", "what different state does `multiplyWithNumberOfRememberMemoryCards` effect?", "what does it mean to pass a `null` DamageType?", all of which required me to dig into the implementation to understand). It is okay for a framework to have some of its own overhead, but it's not helpful to add *more* mental overhead without removing *existing* mental overhead.

> I had to add the Remember Memory card multiplier to WandererActionConfiguration.

> When I was implementing the example, the same thing happened, that always happens, when I design the framework upfront. When actually using the framework, I very rarely have to change its API if something is wrong. The majority of the time, I just fix the implementation of the framework and it works. To me, the card configuration framework is conceptually sound and thus the risk of having to change its API, breaking client code, tends to 0. Of course, it will be extended, but that doesn't break client code.

I think that this usage example actually demonstrates exactly the opposite; in order to use the framework to implement a card with a simple but unique effect, you had to add a new API, new state, and new updates to existing damage calculation methods to the framework, the last of which involves bug risk for every other card using the same damage calculation code, whereas in the original solution, all of the same card-specific unique behavior is encapsulated within the one card.

---

In summary, I do not think there is a significant readability difference between the old and new versions of this example card, and I think there is substantially more mental overhead to understanding the behavior of the new version. I am not willing to merge this based on this example.

HalfTheDeck:
Many things you critizise can be explained to the user by concise javadoc. I don't like to write javadoc when writing classes, because they might still be subject to changes. I like to comment non-private classes, constructors, methods and members when implementation is finished or at least not very likely to be changed significantly anymore. This way you have another implementation step, where you have to think about things, and can identify problems, you missed, and fix them.

Additionally, you asked me in discord to give you my suggestions in small chunks. That's what I did, but in your last comment, you noted, that I had to add a new API. Of course, I wouldn't have had to, if the framework had already been finished. By now, it could be finished, but we wasted so much time doing other things. That is something, that frustrates me.

I didn't give the null parameter value for `damageType` too much attention (I checked the enum for a default value, though, but then I just used null for simplicity). I just extended my framework by the necessary method for handling the Remember Memory tag. A new method without the `damageType` parameter can be added easily.

As for `damage` vs. `baseDamage`, the parameter for the damage in the framework method, that is called in the example, is named `baseDamage`. Javadoc will remove any confusion, that is still left, if any.

The method's name is `multiplyWithRememberMemoryCards` because it might not only be applied to damage values but also block values (or magicNumber values in general). I could add a method named `multiplyDamageWithRememberMemoryCards`, that does the same as `multiplyWithRememberMemoryCards` or maybe even `multiply[Damage]WithNumberOfRememberMemoryCards`. It's just a matter of naming and no general problem of the framework. If you don't like a name, feel free to suggest another one.

If I understand your comment correctly, you are worried, that the framework increases the bug risk for every card, that uses it. Even if there is a bug, it can be fixed in the implementation of the framework. As long as the API is clear, which it will be when it's finished, client code is safe.

The problem with weak or vulnerable is the same, when the framework is not used. I don't know if the value of `damage` reflects those conditions. Maybe the example implementation is even wrong in that regard. But again, that can be fixed in the framework without changing its (public) API. Applying player weakness and monster vulnerability to the calculated damage is the default. If a card is to ignore a condition, there will be a method in the configuration class (e.g. ignoreSourceWeakness() where 'source' is probably always the player).

Last but not least, I always said, that using the framework is optional. I want to use it when i implement cards. No one has to use it. Maybe you're afraid, that someone uses the framework in its current state after you merge the PR. We could mark the classes as `@Deprecated` until it's finished or at least ready to use.

I think the framework can do an excellent job removing any confusion regarding the meaning and context of the different card member values in various states. As my example implementation shows, you just declare what the card does. Default behavior (like applying monster vulnerability) is not written in the cards' texts, either.

dbjorge:
> Many things you critizise can be explained to the user by concise javadoc.

This is not different from it being possible to write documentation explaining the existing system.

> Additionally, you asked me in discord to give you my suggestions in small chunks. That's what I did, but in your last comment, you noted, that I had to add a new API. Of course, I wouldn't have had to, if the framework had already been finished. By now, it could be finished, but we wasted so much time doing other things. That is something, that frustrates me.

I'm not pushing back because the initial version of the framework did not contain the functionality for all cards; I am much happier that you sent a PR with it in that stage than I'd have been with a huge PR with a more complete implementation. I am pushing back because my expectation, based on now having implemented more than half the Wanderer's cards, is that **many** cards would require a similar sort of framework update, because **many** cards have some sort of comparable interesting/unique behavior. I would much prefer that behavior which is unique to a card (like in this example) be encapsulated in that one specific card, rather than being placed in a common class shared by all cards.

> The method's name is multiplyWithRememberMemoryCards because it might not only be applied to damage values but also block values (or magicNumber values in general).

This is, again, a solution in search of a problem. There is no effect that requires this.

> If I understand your comment correctly, you are worried, that the framework increases the bug risk for every card, that uses it. Even if there is a bug, it can be fixed in the implementation of the framework. As long as the API is clear, which it will be when it's finished, client code is safe.

The equivalent client code is *already safe* from bugs in new unique card effects like this *without* the framework because the existing solution encapsulates that unique effect behavior. I would rather prevent bugs from being possible in the first place than fix them.

> The problem with weak or vulnerable is the same, when the framework is not used. I don't know if the value of `damage` reflects those conditions.

Bluntly, I think you would be better served by trying to understand the underlying APIs you're abstracting *before* starting to implement your abstraction layer.

> Last but not least, I always said, that using the framework is optional.

I am not willing to add the maintenance burden of maintaining multiple canonical ways for implementing the same behavior. "There is one contributor that would rather write it this way" is not a good enough justification.

I am sorry, but fundamentally I do not think this abstraction will have a positive impact on the maintainability of the code base. I have spent enough time reviewing, analyzing, and trying this out, and will be closing this PR.

I will not be reviewing further PRs with substantially the same abstraction layer.

HalfTheDeck:
@dbjorge 

Well, I could say a lot in response, but it's not worth the effort, because I don't think anything I could possibly say, would change the way you think. Just one remark: You've clearly shown, that you don't understand what a framework is. That 'solution in search of a problem' bullshit, that you've written multiple times, is proof. A framework is abstract and does not know who uses it. It has an API and does what it says. A good framework is flexible and as such inherently a so-called solution in search of a problem.

As a final remark, I stand by everything I've written in the PRs. Good luck. You will need it!

dbjorge:
I have been civil in my feedback with you, and you have already been warned about acting professionally here; I'm not going to tolerate obscenities and personal attacks in PRs to a side project that I work on in my free time. You are welcome to fork the repository if you'd like, but I'm going to be blocking you from this one.

PR #21, opened 2019-10-20
Manipulation rare: Feast with placeholder art.

Frenoss:


src/main/java/stsjorbsmod/cards/Feast.java
```diff
@@ -0,0 +1,50 @@
...
+
+        tags.add(REMEMBER_MEMORY);
+    }
+
+    @Override
+    public void upgrade() {
+        if (!upgraded) {
+            upgradeName();
+            upgradeMagicNumber(UPGRADE_PLUS_DRAW);
+            //initializeDescription();
```
dbjorge: Is this intended to be commented out?

Frenoss: I'll remove it for now. It seems like most upgrade() overrides don't need to do anything extra to their description, though they do.

src/main/java/stsjorbsmod/cards/Feast.java
```diff
@@ -0,0 +1,49 @@
...
+        magicNumber = baseMagicNumber = DRAW;
+        exhaust = true;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        enqueueAction(new RememberSpecificMemoryAction(new GluttonyMemory(p, false)));
+        enqueueAction(new DrawCardAction(p, magicNumber, false));
+
+        tags.add(REMEMBER_MEMORY);
```
dbjorge: I think the tag addition needs to go in the ctor, not `use` (it exists so `Weight of Memory` can find the card)

Frenoss: I'll change this here. Does that mean MirrorImage, for instance, should be fixed the same way?

dbjorge: Yep, oops

PR #24, opened 2019-10-21
Manipulation rare: Timewalk 

Frenoss:
…and extra dynamic variables for its digity card text.

Time Walk specifies three numbers. I added two extra magic numbers, with playful names. These can be ported for use in a couple other cards (e.g. Corroding Barrier) that currently hard code values, if we want to keep this approach. The alternative is to specify the numbers directly in the strings. It's not bad for values that don't change with upgrades or being modified by other powers.

I waffled whether to import * or the two variables by name in JorbsMod.java. Currently very specific.

The loss and gain of Energy currently has two steps. It could be smooshed into one difference-adjusting step.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -320,5 +320,10 @@
   "stsjorbsmod:Feast": {
     "NAME": "Feast",
     "DESCRIPTION": "Remember stsjorbsmod:Gluttony. NL Draw !M! cards. NL Exhaust."
+  },
+  "stsjorbsmod:TimeWalk": {
+    "NAME": "Time Walk",
+    "DESCRIPTION": "Discard your hand and lose all Block. NL Gain !M! energy and draw !stsjorbsmod:UrMagic! cards. NL Gain !stsjorbsmod:MetaMagic! Weak. NL Exhaust.",
```
Frenoss: Should "energy" be capitalized? I think the main classes have unique symbols for theirs, so they don't offer an exact comparison.

dbjorge: We actually have the ability to use the energy symbol also! The placeholder you want for it is `[E]` (see `Channel`'s text for an example)

src/main/java/stsjorbsmod/cards/TimeWalk.java
```diff
@@ -0,0 +1,59 @@
...
+        magicNumber = baseMagicNumber = ENERGY;
+        urMagicNumber = baseUrMagicNumber = DRAW;
+        metaMagicNumber = baseMetaMagicNumber = WEAK;
+        exhaust = true;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        enqueueAction(new LoseEnergyAction(EnergyPanel.getCurrentEnergy()));
+        enqueueAction(new DiscardAction(p, p, 10, false));
```
dbjorge: BaseMod provides a `BaseMod.MAX_HAND_SIZE` to use instead of hardcoding 10 (improves compat with other mods)

src/main/java/stsjorbsmod/JorbsMod.java
```diff
@@ -295,6 +297,9 @@ public void receiveEditRelics() {
     public void receiveEditCards() {
         logger.info("Adding cards");
 
+        BaseMod.addDynamicVariable(new UrMagicNumber());
```
dbjorge: This is way better than the hardcoding I'd done in a few cases!

src/main/java/stsjorbsmod/cards/TimeWalk.java
Frenoss: Thanks. I had a mental note to look for something like that and apparently lost it.

PR #25, opened 2019-10-21
Manipulation rare: Ivory Tower

Frenoss:
I added a helper action to handle the selection and exhaustion of curse/status cards.

It looked okay in testing, but I definitely didn't dive into what the duration stuff is about. Maybe in a hand of many status or curse cards the work should split across ticks. I'm hoping you know more as a default guideline.


src/main/java/stsjorbsmod/actions/ExhaustNegativeCardAction.java
```diff
@@ -0,0 +1,42 @@
...
+    public ExhaustNegativeCardAction(AbstractCreature source, CardGroup pile) {
+        this.setValues(AbstractDungeon.player, source, 1);
+        this.actionType = ActionType.EXHAUST;
+        this.duration = Settings.ACTION_DUR_FASTER;
+        this.pile = pile;
+    }
+
+    @Override
+    public void update() {
+        ArrayList<AbstractCard> toExhaust = new ArrayList<AbstractCard>();
```
dbjorge: Generally, an `Action`'s update method gets called once per game-update-tick until either its duration has expired (via `tickDuration()` calls that each decrement it by game tick durations) or its update method sets `isDone = true`. Usually, actions will either be:

* "instant": they don't set `this.duration` in their ctor, and their update method just does its thing and unconditionally sets `isDone=true` - usually you see this for actions that work by queuing other action(s) that are non-instant, or
* "non-instant" (they set `this.duration = this.startingDuration = DUR_WHATEVER` in their ctor, they call `this.tickDuration()` unconditionally at the end of their update method, and the actual work is hidden behind a `if (this.duration == this.startingDuration) { /* do work */ }` clause so it only happens once.

This is working like a non-instant action but isn't following the usual pattern; I'd expect it to work fine anyway because its work happens to be idempotent, but I think it would be good to follow the usual pattern anyway (like `ExhaustSpecificCardAction` does) for consistency.

src/main/java/stsjorbsmod/actions/ExhaustNegativeCardAction.java
```diff
@@ -0,0 +1,42 @@
...
+import com.megacrit.cardcrawl.core.Settings;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+
+import java.util.ArrayList;
+
+// Based on ExhaustSpecificCardAction
+public class ExhaustNegativeCardAction extends AbstractGameAction {
+    private CardGroup pile;
+
+    public ExhaustNegativeCardAction(AbstractCreature source, CardGroup pile) {
```
dbjorge: I was initially confused by the word "Negative" here, I read it as the "with a value < 0" form rather than the "with a deterimental effect" form. Maybe it would be less ambiguous to instead use `ExhaustCardsMatchingPredicateAction` and have `IvoryTower` pass a `Predicate<AbstractCard>` of `c -> c.type == CardType.CURSE || c.type == CardType.STATUS`?

Frenoss: Great idea; updated.

src/main/java/stsjorbsmod/actions/ExhaustNegativeCardAction.java
Frenoss: Ewww, duration is a float. I can follow the base game's pattern, which should work because those variables are initialized to be equal, but that bit of it is ugly.

Frenoss: Do you have any preference for use of "this."? My slow drift is to not use it except in constructors.

dbjorge: I don't really have a preference on "this", since IntelliJ does a reasonable job of coloring locals/fields differently and warning on unused assignments.

PR #26, opened 2019-10-21
Manipulation rare: Sharpened Mind

Frenoss:


src/main/java/stsjorbsmod/cards/SharpenedMind.java
```diff
@@ -0,0 +1,47 @@
...
+import stsjorbsmod.JorbsMod;
+import stsjorbsmod.characters.Wanderer;
+import stsjorbsmod.powers.SharpenedMindPower;
+
+import static stsjorbsmod.JorbsMod.makeCardPath;
+import static stsjorbsmod.characters.Wanderer.Enums.REMEMBER_MEMORY;
+
+// [Power, unstackable]
+// 1(0): At the start of your turn, remember Patience. NL At the end of your turn, remember Diligence.
+// Note: this implementation means the only effect from Diligence is the card draw. By the time Diligence
```
dbjorge: The implementation looks great to me, this note is the only thing that would maybe be good to change; my presumption is that the intention is for Diligence's retain effect to work here. Would the right ordering happen if you switched L49 of the power to use `addToTop` instead of `addToBottom`?

Frenoss: Unfortunately, `addToTop is still after any `atEndOfTurn would be invoked. I added a timing variable and work-around in the Diligence memory itself. Check if this isn't too hacky.

The timing itself should be safe. Powers all have their `atEndOfTurn methods invoked before the hand discard action is queued.

PR #29, opened 2019-10-23
Manipulation rare: Dimension Door

Frenoss:
So, I went with a slightly different approach. I think it's semantically shakier but more resilient by only using postfix patches.

I took the stance that we don't actually care to say we are flying, when Dimension Door is active. Instead, we extend the basic notion of "connected".

Another part is invoking a new SmokeBomb instead of copying its implementation. I wanted the benefit of getting any implementation changes. I didn't observe any extraneous effects, but I don't have a great feel for what's more stable long-term.

Oh, and the card text is very long. It uses all of the explanations from Smoke Bomb and Winged Greaves to be as clear as possible. Running to seven lines might just be less clear, though.

(And, would all of these comments be better in the commit message?)


src/main/java/stsjorbsmod/cards/DimensionDoor.java
```diff
@@ -0,0 +1,50 @@
...
+        super(ID, IMG, COST, TYPE, COLOR, RARITY, TARGET);
+        this.upgraded = true;
+        this.purgeOnUse = true;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        activatedMapTravel = true;
+
+        p.masterDeck.removeCard(this.cardID);
```
dbjorge: You want to remove by uuid, not cardID; 2 different Dimension Door cards will have the same cardID but different uuids, which has particularly bad results in combination with effects that create temporary/duplicate copies of cards. `StSLib` provides a "search for equivalent uuid" helper for this (`StSLib.getMasterDeckEquivalent`), but note that the temporary/duplicate card thing means we'll want to check its return value for null.

`Wish_Wanderer` already does exactly this to implement the same type of "destroy card on use" effect, maybe it would be good to lift that helper to `CustomJorbsModCard` and reuse it.

src/main/java/stsjorbsmod/cards/DimensionDoor.java
```diff
@@ -0,0 +1,50 @@
...
+        return activatedMapTravel;
+    }
+
+    public static void deactivateMapTravel() {
+        activatedMapTravel = false;
+    }
+
+    public DimensionDoor() {
+        super(ID, IMG, COST, TYPE, COLOR, RARITY, TARGET);
+        this.upgraded = true;
```
dbjorge: It looks a little off for it to start out upgraded, I think, though I understand that it's awkward to come up with a good upgrade effect. Maybe upgrade to "you can choose any floor from the next 2 levels, rather than the next 1 level"?

src/main/java/stsjorbsmod/cards/DimensionDoor.java
```diff
@@ -0,0 +1,50 @@
...
+        this.purgeOnUse = true;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        activatedMapTravel = true;
+
+        p.masterDeck.removeCard(this.cardID);
+
+        (new SmokeBomb()).use(p);
```
dbjorge: I think the bits of `SmokeBomb.use` we'd need is small enough that I'd rather repeat it than worry about whether there's any exciting interactions with competing SpirePatches, especially since it'd be nice to have a unique VFXAction effect for it at some point.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -317,6 +317,10 @@
     "NAME": "Banish",
     "DESCRIPTION": "Enemy takes no action and no damage for !M! turns. Exhaust."
   },
+  "stsjorbsmod:DimensionDoor": {
+    "NAME": "Dimension Door",
+    "DESCRIPTION": "Escape from a non-boss combat and receive no rewards. NL If you escape, you may ignore paths when choosing the next room to travel to. NL Destroy this card."
```
dbjorge: I think this is still allowing escape from boss combats as-is

src/main/java/stsjorbsmod/patches/DimensionDoorPatch.java
```diff
@@ -0,0 +1,55 @@
...
+    {
+        @SpirePostfixPatch
+        public static boolean patch(boolean __result, MapRoomNode __this, MapRoomNode __node)
+        {
+            if (__result || !DimensionDoor.isMapTravelActive()) {
+                return __result;
+            }
+
+            // Any room on the next level is actually reachable.
+            Iterator edgeIterator = __this.getEdges().iterator();
```
dbjorge: The `.iterator()` stuff is a common decompiler output, but I think it reads a littler clearer if you rewrite it as

```java
for (MapEdge edge : __this.getEdges()) {
    if (__node.y == edge.dstY) {
        return true;
    }
}
```

dbjorge: In this case, the `stream()` based version is also pretty readable

```java
return __this.getEdges().stream().anyMatch(e -> __node.y == e.dstY);
```

src/main/java/stsjorbsmod/cards/DimensionDoor.java
Frenoss: This turns out to be a fun effect. So many available locations! The default interaction with the boss room works well. It can't be reached in this manner, so a person won't accidentally or intentionally skip the last campfire.

src/main/java/stsjorbsmod/cards/DimensionDoor.java
Frenoss: While I was thinking the code could be lifted at any point new VFX were ready, I really like your point about other patches.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
Frenoss: Oopsie.

src/main/java/stsjorbsmod/patches/DimensionDoorPatch.java
Frenoss: Thank you for connecting those dots! I had seen a use of stream in the Java docs but dismissed it as an IO thing. Hmm, I see now I went with a different formatting. It feels okay since functions weren't chained, but I understand if preferred to be in-line with your example style.

PR #32, opened 2019-10-24
Add extra check for ALL_ENEMY signifier

Frenoss:


dbjorge:
Ended up going with a slightly bigger fix that also corrects the null check (it should have been on the parameter `target`, not `card.target` and fixes behavior in cases that don't target enemies at all (https://github.com/dbjorge/jorbs-spire-mod/commit/1158714db419ae94bc326715fee5f6e42525c050)

PR #34, opened 2019-10-24
Move wanderer cards to a sub-directory

Frenoss:
Hmm, I'm worried that it can't automatically merge. I think now I should have split into two commits, one purely for rename, and one for cleanup of imports and packages. Let me know if doing so will be better from your side.

This does nothing with resources. I think it would make sense to have a different card strings file per class. Keep images as they are for now, since they follow their own hierarchy.


Frenoss:
Oh, I missed that the queue update went through. I will redo this so you don't have to go through that.

PR #38, opened 2019-10-27
Fix v1.0/v2.0 incompatibilities

mikeynap:
I wanted to play the mod on the beta branch (and take my chances with any runtime inconsistencies). There were a few super trivial things to make it at least compile.

This is just a start obviously. I'll do some beta against on the beta branch to see if I can find any other issues. 

~ thefunkiestchicken


mikeynap:
Shoot... because non-beta does not have this method, @override doesn't work.

hmmm....

Issue #52, opened 2019-10-28
Artifact should not block SnappedPower


dbjorge:
Probably best to add a new power type (maybe SPECIAL) as an alternative to BUFF/DEBUFF, since we'll want SnappedPower to be immune to both debuff-affecting things like artifact but also buff-affecting things like Amnesia

wang429:
Related, current Amnesia order cleanses SnappedPower. RemoveAllPowersAction removes _all_ powers and is done after SnapAction. Either new "RemoveNonSpecialPowersAction" or patch RemoveAllPowersAction to ignore SPECIAL power types.

dbjorge:
Yep, Amnesia will need an update. I think I'd probably lean towards making a new `RemovePowersMatchingPredicateAction` (similar to what we did for `ExhaustCardsMatchingPredicateAction`) and let `Amnesia` pass in its predicate for `p -> p.type == PowerType.BUFF || p.type == PowerType.DEBUFF`

wang429:
Oh, Amnesia is also a curse now.

dbjorge:
Fixed by #96

PR #96, opened 2019-11-04
SnappedPower should be unblockable uncleansable

wang429:
Artifacts doesn't block SnappedPower from applying anymore
Amnesia doesn't cleanse SnappedPower anymore
Amnesia is now an unplayable curse. card effects occur at the end of turn.
Amnesia text order to betters matches order of card actions


wang429:
Addresses #52

Issue #58, opened 2019-10-28
Wrath damage upgrades reset when quitting and reloading a save


dbjorge:
User reported issue with Magic Missiles specifically

dbjorge:
See https://github.com/daviscook477/BaseMod/wiki/Custom-Savable for how to implement fixing this

dbjorge:
Fixed by #106

Issue #36, opened 2019-10-27
Card titles should indicate Wrath metaupgrades


dbjorge:
It would be nice to have some sort of indicator in the card title when Wrath causes a damage increase to a card. Current proposal would be to update the card render() to look for the presence of wrath upgrades and, if present, render an overlay image that adds a wrath icon to the top-right corner and render the increase count on top of that, similar to how rendering energy cost works.

![wrath_card_title_concept](https://user-images.githubusercontent.com/376284/68361651-8192b480-00d9-11ea-8f72-fef661260a93.png)

Note that Wrath can upgrade base game colorless cards! So this will need to be implemented with `SpirePatch`es on `AbstractCard`, **not** with overrides in `CustomJorbsModCard`.

dbjorge:
Implemented in 3fd9aa2

Frenoss:
Maybe move the text a tiny bit to the right. What say ye?
![image](https://user-images.githubusercontent.com/6187622/68461536-48347480-01bf-11ea-9604-7333c6114fa2.png)

dbjorge:
Yes to both

dbjorge:
![image](https://user-images.githubusercontent.com/376284/68497006-6ecacd80-0208-11ea-8c3e-cb27abedd9a6.png)

Moved text slightly right, still need to do the portrait-view bit

dbjorge:
![image](https://user-images.githubusercontent.com/376284/68500235-041d9000-0210-11ea-8970-3b23e8edaf73.png)

fixed as of 2dcb32a

PR #106, opened 2019-11-05
Persist Wrath modifiers across play sessions

Frenoss:
Cards save their ID, number of upgrades, and their misc field by
default. To save how Wrath modifies a card's base damage, we
implement the CustomSavable interface on all of our cards and
introduce a new class to track persistent changes. Currently the only
value is how much Wrath modifies a card. Persistent changes are
re-applied to cards upon loading.

REVIEW: The addition of the save class seems like extra work for what's just one field. When I originally had it implemented as an Integer, I was seeing odd behavior, where the value seemed to be lost over time. I was wondering if something about a common class like Integer caused problems with deserializing the JSON in the mainline code. I never root caused the problem, but after changing to a unique class I haven't been able to repro the behavior.

### DIFF
--- a/src/main/java/stsjorbsmod/memories/WrathMemory.java
+++ b/src/main/java/stsjorbsmod/memories/WrathMemory.java
@@ -25,7 +25,7 @@ public class WrathMemory extends AbstractMemory {
 
     public static void reapplyToDeck(CardGroup deck) {
         for (AbstractCard card : deck.group) {
-            final int upgradeCount = WrathField.upgradeCount.get(card);
+            final int upgradeCount = WrathField.effectCount.get(card);
             if (!isUpgradeCandidate(card) && (upgradeCount > 0)) {
                 JorbsMod.logger.error("Wrath upgrade count modified for an ineligible card");
             }
@@ -95,18 +95,14 @@ public class WrathMemory extends AbstractMemory {
         AbstractCard masterCard = StSLib.getMasterDeckEquivalent(card);
         if (masterCard != null) {
             masterCard.baseDamage += DAMAGE_INCREASE_PER_KILL;
+            WrathField.effectCount.set(masterCard, WrathField.effectCount.get(masterCard) + 1);
             masterCard.superFlash();
             cardToShowForVfx = masterCard;
-
-            // Because it is the master deck that's saved, we need to track
-            // the card's deviation from base for potential future loads.
-            int count = WrathField.upgradeCount.get(masterCard);
-            WrathField.upgradeCount.set(masterCard, count + 1);
         }
 
         for (AbstractCard instance : GetAllInBattleInstances.get(card.uuid)) {
-            instance.isDamageModified = true;
             instance.baseDamage += DAMAGE_INCREASE_PER_KILL;
+            WrathField.effectCount.set(instance, WrathField.effectCount.get(instance) + 1);
             instance.applyPowers();
         }
 
--- a/src/main/java/stsjorbsmod/memories/WrathMemory.java
+++ b/src/main/java/stsjorbsmod/memories/WrathMemory.java
@@ -3,7 +3,6 @@ package stsjorbsmod.memories;
 import com.evacipated.cardcrawl.mod.stslib.StSLib;
 import com.megacrit.cardcrawl.cards.AbstractCard;
 import com.megacrit.cardcrawl.cards.AbstractCard.CardType;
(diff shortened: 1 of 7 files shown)


Frenoss:
Resolves #58.

dbjorge:
I don't think we can use the strategy of making the `CustomJorbsModCard` implement the savable because Wrath can also upgrade base game colorless cards that don't use our card base.

I think instead it would probably be better to use a [SpireField](https://github.com/kiooeht/ModTheSpire/wiki/SpireField) on `AbstractCard` to store the number of upgrades (we'll want this anyway for the sake of being able to display it on card titles), and then use a mod-global `BaseMod.addSaveField(String key, CustomSavableRaw saveField)` to add logic for storing a "uid -> increaseCount" map.

Frenoss:
Whoa, nifty. Thanks for the pointer.

When I go to do that, what's the favored way for prettifying the commit history? Submit a new pull request; rebase previous commits away; keep a record of all the commits anyway?

dbjorge:
I like to use squash-merges for PRs, so the history on master is going to end up clean looking after the PR is merged regardless. I usually prefer to avoid rebasing and force-pushing for PRs that have already been submitted because rewriting history of the branch is a bit awkward for anyone pulling down your changes to test locally, so I would say prefer to just keep a record of all commits, or if you're truly starting over from scratch and think the extra history would confuse PR reviewers, start a new PR

dbjorge:
Note that I'm not sure the order of loading is guaranteed for the custom savable; I'm not sure if you'll be able to do it the easy way (just enumerate masterDeck during load assuming all the cards will already be initialized there) or something more complicated with separate "load save file" and "update cards later" logic

src/main/java/stsjorbsmod/patches/WrathField.java
```diff
@@ -0,0 +1,14 @@
...
+import com.evacipated.cardcrawl.modthespire.lib.SpirePatch;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+
+@SpirePatch(
+        clz = AbstractCard.class,
+        method = SpirePatch.CLASS
+)
+public class WrathField {
+    @SuppressWarnings("unchecked")
+    public static SpireField<Integer> upgradeCount = new SpireField(() -> 0);
```
dbjorge: nit: the word `upgrade` is a little overloaded as applied to cards, and the field name here is what actually gets injected as the field name on the card, so it will show up in debugging. Maybe `wrathDamageIncrease` would be more clear?

src/main/java/stsjorbsmod/memories/WrathMemory.java
```diff
@@ -86,9 +97,15 @@ private void permanentlyIncreaseCardDamage(AbstractCard card) {
...
             cardToShowForVfx = masterCard;
+
+            // Because it is the master deck that's saved, we need to track
+            // the card's deviation from base for potential future loads.
+            int count = WrathField.upgradeCount.get(masterCard);
+            WrathField.upgradeCount.set(masterCard, count + 1);
         }
 
         for (AbstractCard instance : GetAllInBattleInstances.get(card.uuid)) {
+            instance.isDamageModified = true;
```
dbjorge: I don't think we want to set this; it recolors the number on the card in way that usually implies that the instance is being modified in a non-permanent way. #36 tracks updating the title of the card to note the wrath upgrade; I think that would be a better way to indicate it to the user.

Instead, consider setting the WrathField here on the instances as well; that'll probably prevent some head scratching if someone else ends up picking up #36 :)

dbjorge: More subtle interaction with #36: we probably want to patch `AbstractCard.makeStatEquivalentCopy()` to copy over the wrath field (probably our ur/meta numbers, too)

src/main/java/stsjorbsmod/patches/WrathField.java
Frenoss: I want to indicate the number of times Wrath affected it instead of the damage increase. Maybe `wrathDamageInstances` or `wrathApplicationCount`. It will be something.

src/main/java/stsjorbsmod/memories/WrathMemory.java
Frenoss: Drat, I like to finish the thing before bed, but pushing commits after 1am comes with a certain risk. The isDamageModified shouldn't be there at all.

Thanks for the pointer on `makeStatEquivalentCopy()`.

PR #108, opened 2019-11-06
Calculating damages appropriately for ChainLightning.

### LINKED ISSUE #104: Chain Lightning damage
Chain Lightning seems to only deal half damage to other enemies after targeting Byrd in a fight.

stevematney:
This ensures damage gets calculated correctly for all enemies. Resolves #104


src/main/java/stsjorbsmod/cards/wanderer/ChainLightning.java
```diff
@@ -35,8 +37,42 @@ public ChainLightning() {
...
+            return monsterDamage;
+        }).mapToInt(i -> i).toArray();
+
+        damageMultiplier = 0;
+
+        return damagesArray;
+    }
+
+    private int getDamage(AbstractMonster monster, int multiplier) {
+        this.baseDamage = DAMAGE + (DAMAGE_PLUS_PER_HOP * multiplier);
```
dbjorge: I think this should use `magicNumber` in place of `DAMAGE_PLUS_PER_HOP`, to enable the upgrade effect

stevematney: Oh. Good point! I'll add that. I'm also doing some refactoring to make sure we always hit and add effects for the correct enemies. (That is, even when the damage is 0 but we _should_ be hitting the enemy.)

dbjorge: Setting `baseDamage` like this is going to have the side effect of ignoring Wrath bonuses. Instead of this line, you can add an override for `calculateBonusBaseDamage` based on `damageMultiplier` to get a similar effect that's more wrath-friendly, eg:

```java
@Override
    return magicNumber * damageMultiplier;
}
```

src/main/java/stsjorbsmod/actions/ChainLightningAction.java
```diff
@@ -32,27 +32,18 @@ public ChainLightningAction(AbstractCreature owner, AbstractMonster initialTarge
...
-
-            nextTargetIndex = AbstractDungeon.cardRandomRng.random(0, remainingTargets.size() - 1);
+        int currentDamage = 0;
+        AbstractMonster currentTarget;
+        for (int i = 0; i < allTargets.size(); i++) {
+            currentTarget = allTargets.get(i);
+            currentDamage = this.damage[i];
+            if (currentDamage <= 0) continue;
+
+            AbstractDungeon.actionManager.addToBottom(
```
dbjorge: I think at this point it's probably not necessary to have a separate action at all; I think we could probably do like ThunderClap does and just queue the `DamageAction`s and `VFXAction`s from the card directly and get rid of the intermediate damage/order arrays

stevematney: Yeah! That simplifies things a bit, for sure.

src/main/java/stsjorbsmod/cards/wanderer/ChainLightning.java
```diff
@@ -27,16 +32,54 @@
...
+        this.damageMultiplier = 0;
+        for (AbstractMonster monster: targets) {
+            this.calculateCardDamage(monster);
+            AbstractDungeon.actionManager.addToBottom(new DamageAction(
+                    monster,
+                    new DamageInfo(p, this.damage, DamageInfo.DamageType.NORMAL),
+                    AttackEffect.NONE));
+            addLightningEffect(monster, this.damageMultiplier);
+            this.damageMultiplier += 1;
+        }
```
dbjorge: This refactor looks much better! I think the "reseting `damageMultiplier` at the end of use" that was here previously needs to be put back, though, otherwise I think the highest `damageMultiplier` from the last time it was played will get calculated into the damage displayed on the card description the next time through the deck

stevematney: Good point! I’ll add that back tomorrow and we can get this merged.

dbjorge: I'll go ahead and add it so I can include it in a release tonight. Thanks for the work!

Issue #46, opened 2019-10-27
After skipping a floor with Dimension Door, floor number is wrong


dbjorge:

dbjorge:
It was pointed out in chat that the base Portal event works this way too, so maybe this is fine

dbjorge:
Probably obsoleted by #44

dbjorge:
Fixed by #110

PR #110, opened 2019-11-06
Replace Dimension Door card with potion

### LINKED ISSUE #44: Dimension Door as a potion
Maybe Dimension Door should be a potion rather than a card. Needs further discussion

Frenoss:
Remove the Wanderer's card Dimension Door. Add a new potion for the
Scroll of Dimension Door. The behavior for travel is implemented.
Currently, the potion has a limitation that it must be used during an
encounter before seeing the map.

Resolves #44 and #46.

I think because it works enough to let people play with and balance the effects, it's worth putting in. Stuff still to do:
- It's not friendly to use, in that players would want to click it when looking at the map. That's currently not possible.
- Remove logic for extra travel range if that's never going to be used.
- It needs to be updated to look like a scroll. I haven't looked at all into laying the groundwork for those textures.


dbjorge:
Pushed an update with a hack that allows it to be used when looking at the map (making it extend FruitJuice) - the patches required to emulate it were worse. Agreed that we don't need the scroll art to merge. Thanks!

Frenoss:
I really like this idea. I had discarded it, but I think I should change my metric to be closer to: what’s less hack code. This is less than the probably three patches, and two of them might have been nasty, raw ones, for making it work otherwise.

I guess in the world of modding StS, everything’s a hack.

One thing I hadn’t checked, because I didn’t have this working, does the reachability check update if you drink it on the map? I’ll pull master and give it a try.

________________________________

Frenoss:
Oh wow, it just works.

________________________________
From: Aaron Schrader <aaronsch@hotmail.com>
Sent: Thursday, November 7, 2019 11:48:04 PM
To: dbjorge/jorbs-spire-mod <reply@reply.github.com>; dbjorge/jorbs-spire-mod <jorbs-spire-mod@noreply.github.com>
Cc: Author <author@noreply.github.com>
Subject: RE: [dbjorge/jorbs-spire-mod] Replace Dimension Door card with potion (#110)


I really like this idea. I had discarded it, but I think I should change my metric to be closer to: what’s less hack code. This is less than the probably three patches, and two of them might have been nasty, raw ones, for making it work otherwise.



I guess in the world of modding StS, everything’s a hack.



One thing I hadn’t checked, because I didn’t have this working, does the reachability check update if you drink it on the map? I’ll pull master and give it a try.



________________________________

PR #119, opened 2019-11-09
Update Gluttony memory

Frenoss:
Reduce Max HP gain from 3 to 2.
Let player pick discard instead of random.
Piece together updated FRA translation from other cards.


dbjorge:
I just went ahead and removed the machine french translations since they had already become substantially out of date and will get worse as the balance pass continues; we can regenerate it once the mod is more stable

Frenoss:
Hah, okay, I was about to ask why we were going with such a differing take on French from the main game.

Sorry I didn’t submit these two when I had them done yesterday. I’m not sure why that was the case.

________________________________

dbjorge:
They're almost-strictly from google translate, I added it during a charity stream that was using "play in a language I don't know how to speak" as a donation incentive. I don't think it's worth maintaining until the mod is more stable.

PR #122, opened 2019-11-09
Update Firebolt

Frenoss:


src/main/java/stsjorbsmod/cards/wanderer/Firebolt.java
```diff
@@ -62,7 +63,7 @@ public void use(AbstractPlayer p, AbstractMonster m) {
     public void upgrade() {
         if (!upgraded) {
             upgradeName();
-            upgradeDescription();
+            upgradeMagicNumber(UPGRADE_PLUS_PER_CLARITY);
```
dbjorge: Needs to keep an `initializeDescription()` still; that's what triggers the `!M!` placeholder to update immediately in the description

src/main/java/stsjorbsmod/cards/wanderer/Firebolt.java
```diff
@@ -30,7 +30,8 @@
 
     private static final int COST = 1;
     private static final int DAMAGE = 8;
-    private static final int AMOUNT_PER_CLARITY = 2;
+    private static final int AMOUNT_PER_CLARITY = 1;
+    private static final int UPGRADE_PLUS_PER_CLARITY = 1;
```
dbjorge: I think this needs a more significant update to remove `calculateBonusBaseDamage` and remove the `if(upgraded)` conditional from the burning effect in `use()`

src/main/java/stsjorbsmod/cards/wanderer/Firebolt.java
Frenoss: Hmm, in the few different upgrade views I've tried, I always see the upgraded number. Do you have a repro?

src/main/java/stsjorbsmod/cards/wanderer/Firebolt.java
Frenoss: Yes it does. Oy.

src/main/java/stsjorbsmod/cards/wanderer/Firebolt.java
dbjorge: I don't remember offhand, sorry, I just remember having to fix this before for other cards; I think it was something obscure like "when an upgraded card shows up in a shop" or "when the tongs relic upgrades a card" or something

Frenoss: I'm sorry to be stuck on this, but I really do wonder if something's going on there or what. I have now tried all of the relics that mention upgrading attacks, and they all look to have worked correctly. I think I could reduce the card pool and try out the merchant problem.

I guess a big part of my reluctance is that the base game's cards don't modify their description unless text changes. That makes it seem like the right way to do it.

dbjorge: If the base game doesn't do it, I'm probably misremembering. Would like to update the rest of the cards to be consistent, though, if it's missing in just one spot it looks like a bug.

PR #123, opened 2019-11-09
Update Hibernate

Frenoss:


Frenoss:
Oops, I might have forgotten to copy the In Progress symbol to the sheet. It looks like it's already done.

PR #126, opened 2019-11-09
Minor refactoring around Chain Lightning's use.

stevematney:
I was thinking about this code, and thought it could be streamlined just a little bit. If this isn't the preferred style or is drastically less performant for some reason (my measurements were 0-1ms for the loop before the changes and after), I don't mind if we just toss this. Just thought I'd put it out there as a little simplification.

### DIFF
--- a/src/main/java/stsjorbsmod/cards/wanderer/ChainLightning.java
+++ b/src/main/java/stsjorbsmod/cards/wanderer/ChainLightning.java
@@ -14,7 +14,7 @@ import stsjorbsmod.cards.CustomJorbsModCard;
 import stsjorbsmod.characters.Wanderer;
 
 import java.util.ArrayList;
-import java.util.stream.Collectors;
+import java.util.Arrays;
 
 import static stsjorbsmod.JorbsMod.makeCardPath;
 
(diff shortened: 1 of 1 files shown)


src/main/java/stsjorbsmod/cards/wanderer/ChainLightning.java
```diff
@@ -47,34 +47,33 @@ protected int calculateBonusBaseDamage() {
...
-            if (newTargets.isEmpty()) break;
-            nextTargetIndex = AbstractDungeon.cardRandomRng.random(0, newTargets.size() - 1);
-        }
+        ArrayList<AbstractMonster> finalTargets = new ArrayList<>(Arrays.asList(initialTarget));
+        targets.stream()
+                .filter(t -> !(t.halfDead || t.isDying || t.isEscaping || t == initialTarget))
+                .forEach(t -> {
+                    int targetIndex = AbstractDungeon.cardRandomRng.random(0, finalTargets.size() - 1);
+                    // Never add anything other than initialTarget at 0
+                    targetIndex = Math.max(1, targetIndex);
```
dbjorge: This makes it non-uniformly random (ie, it is twice as likely for the last monster to be second in the chain than it should be)

src/main/java/stsjorbsmod/cards/wanderer/ChainLightning.java
```diff
@@ -47,34 +47,33 @@ protected int calculateBonusBaseDamage() {
...
-        int nextTargetIndex = newTargets.indexOf(initialTarget);
-        while(true) {
-            finalTargets.add(newTargets.remove(nextTargetIndex));
-            if (newTargets.isEmpty()) break;
-            nextTargetIndex = AbstractDungeon.cardRandomRng.random(0, newTargets.size() - 1);
-        }
+        ArrayList<AbstractMonster> finalTargets = new ArrayList<>(Arrays.asList(initialTarget));
+        targets.stream()
+                .filter(t -> !(t.halfDead || t.isDying || t.isEscaping || t == initialTarget))
+                .forEach(t -> {
```
dbjorge: If you're looking to streamline this, I think even better would probably be to not implement our own order-shuffling at all; I think you could just form the candidate list and then call `Collections.shuffle(chainCandidates, AbstractDungeon.cardRandomRng.random);`

stevematney: I like that a lot better!

PR #132, opened 2019-11-11
Fix card interaction with Nilry's Codex

### LINKED ISSUE #70: Nilary's Codex should not generate cards that grant metaupgrades
* Anything that grants greed/gluttony/wrath
* Anything that heals you

Frenoss:
A number of random card generation methods eliminate cards with
healing effects from the eligible set. We honor the letter and the
spirit of that filtering by tagging our direct healing cards and,
additionally, eliminating cards with a new tag indicating any other
persistent positive effects from being generated as well.

Future cards must be careful to use these tags in order to
indicate when they are not eligible to be generated as a random card.

Resolves #70.

NOTE: I took two stylistic compressions, and I'd like to be sure they seem right to other eyes.
* Patch all of the relevant functions in one file, using inner static classes.
* Use a unified locator for all patches. If a second locator had been needed, I would have given it a more specific name at that point.

NOTE: The colorless card patch isn't necessary at the moment. The more I think about it, the more I think I should remove it.

### DIFF
--- a/src/main/java/stsjorbsmod/patches/FilterRandomCardGenerationPatch.java
+++ b/src/main/java/stsjorbsmod/patches/FilterRandomCardGenerationPatch.java
@@ -26,8 +26,8 @@ public class FilterRandomCardGenerationPatch {
                 localvars = "list"
         )
         @SuppressWarnings("unchecked")
-        public static void patch(@ByRef ArrayList[] list) {
-            RemovePersistentPositiveEffects(list[0]);
+        public static void patch(ArrayList list) {
+            RemovePersistentPositiveEffects(list);
         }
     }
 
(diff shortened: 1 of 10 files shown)


src/main/java/stsjorbsmod/patches/FilterRandomCardGenerationPatch.java
```diff
@@ -0,0 +1,75 @@
...
+            method = "returnTrulyRandomCardInCombat",
+            paramtypez = { }
+    )
+    public static class AbstractDungeon_returnTrulyRandomCardInCombat_1 {
+        @SpireInsertPatch(
+                locator = Locator.class,
+                localvars = "list"
+        )
+        @SuppressWarnings("unchecked")
+        public static void patch(@ByRef ArrayList[] list) {
```
dbjorge: This doesn't need to be byref; you're modifying the contents of the list, not pointing it to a new list.

dbjorge: (same elsewhere)

Frenoss: It's hilarious/sad that I knew that for the helper function.

Frenoss: I just remembered what the kids say: smh.
I don't have an emoji lexicon to remember anything from.

dbjorge:
This looks great besides the 1 minor comment, I like the approach you took. I think I lean towards keeping the colorless patch.

PR #134, opened 2019-11-11
Standardize upgrade to always upgradeDescription

Frenoss:
This creates consistency for how cards write their upgrade() overloads.
If a different string isn't found, then the rawDescription is not reset
nor is the description re-initialized.

Some minor card text changes as part of review.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -51,7 +51,7 @@
   },
   "stsjorbsmod:Counterspell": {
     "NAME": "Counterspell",
-    "DESCRIPTION": "Gain !B! Block. NL Gain !M! Artifact if enemy intends to apply a negative effect. NL Exhaust."
+    "DESCRIPTION": "Gain !B! Block. NL Gain !M! Artifact if an enemy intends to apply a negative effect. NL Exhaust."
```
dbjorge: I think the original wording is more correct here, since Counterspell intentionally requires targetting a specific enemy for flavor reasons

Frenoss: That makes sense. I want to look at main game examples to see if they say "the". I'll save that for a different day and revert this for now.

dbjorge: I went ahead and took care of it and the merge conflicts (wanted the change in before I started making any further updates to the card strings file, to avoid further conflicts)

PR #137, opened 2019-11-12
Scorching ray

wang429:
Most of the interactions in the following link. Added VFX in video do not exactly what's been commit.
https://youtu.be/w86K5LqjmC8


wang429:
Thought of it this morning; I can probably get away with using DamageAction and removing DamageAsBurningAction.

PR #148, opened 2019-11-14
Legendary

Frenoss:
_Notes on the expected final design behaviors from comments in JorbsMod_
Use on a card that can only ever enter the deck by special means, and cannot be removed or transformed once present. The card can exhaust but still counts as present.
Interaction notes:
- Legendary cards will never appear when drafting with the Draft mod enabled. In the implementation as is, CardRewardScreen.draftOpen() calls AbstractDungeon.returnRandomCard() to show three cards. We have filtered the reward pools that returnRandomCard() draws from.
- Legendary rare cards won't be added to the deck with Shiny enabled.  In the implementation as is, the player invokes AbstractDungeon.getEachRare() instead of the CardLibrary's version. This again uses the filtered reward pools, whereas the CardLibrary version would use the full set of available cards.
- Any red, green, or blue cards that duplicate a card could pick a Legendary card. These behaviors remain because this mod is not designed to interact with colored cards from the main game. Known cards: Dual Wield, Nightmare
- Living Wall event: if the player has no purgeable cards, they won't be given a chance to upgrade.
- Designer event: the Remove option can be active without any purgeable cards to select from.

Other notes from the Issue comments:
What's missing?
- validation in *transformCard
- raw patches for DollysMirror and FountainOfCurseRemoval
- validation in onRemoveFromMasterDeck.
Basically, most known situations work, and we're missing logging or extra care for when something unanticipated happens (say, our mod in conjunction with another).

### DIFF
--- a/src/main/java/stsjorbsmod/JorbsMod.java
+++ b/src/main/java/stsjorbsmod/JorbsMod.java
@@ -9,8 +9,10 @@ import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.Texture;
 import com.evacipated.cardcrawl.mod.stslib.Keyword;
 import com.evacipated.cardcrawl.modthespire.lib.SpireConfig;
+import com.evacipated.cardcrawl.modthespire.lib.SpireEnum;
 import com.evacipated.cardcrawl.modthespire.lib.SpireInitializer;
 import com.google.gson.Gson;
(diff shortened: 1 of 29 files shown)


dbjorge:
Wow, this is really well done. I think it's reasonable to merge this as-is but leave the issue open to track the remaining few TODOs on it.

PR #149, opened 2019-11-14
Alchemists fire

wang429:
Relic with some silly right click effects


src/main/java/stsjorbsmod/powers/BurningPower.java
```diff
@@ -60,12 +60,20 @@ public void playApplyPowerSfx() {
...
     @Override
     public void updateDescription() {
-        if (amount <= 0) {
+        if (this.amount <= 0) {
             this.description = DESCRIPTIONS[4];
-        } else if (this.owner != null && !this.owner.isPlayer) {
-            this.description = DESCRIPTIONS[3] + this.amount + DESCRIPTIONS[1] + (this.amount - this.amount / 2) + DESCRIPTIONS[2] + DESCRIPTIONS[4];
         } else {
-            this.description = DESCRIPTIONS[0] + this.amount + DESCRIPTIONS[1] + (this.amount - this.amount / 2) + DESCRIPTIONS[2] + DESCRIPTIONS[4];
+            int amountToReduceBy;
```
dbjorge: Maybe extract this amountToReduceBy calculation to a `calculateNextFalloffAmount` or something so it can be reused in BurningLoseHpAction and keep the relic special case to one spot?

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Relic-Strings.json
```diff
@@ -19,5 +19,12 @@
     "DESCRIPTIONS": [
       "Gaining stsjorbsmod:Clarity deals 3 damage to all enemies. NL Gaining 10 stsjorbsmod:Clarity in the same fight deals 100 damage to all enemies."
     ]
+  },
+  "stsjorbsmod:AlchemistsFireRelic": {
+    "NAME": "Alchemist's Fire",
+    "FLAVOR": "Forca Messi! Forca Messi! Messi Animal!",
+    "DESCRIPTIONS": [
+      "Reduces stsjorbsmod:Burning falloff rate to 10% per turn"
```
dbjorge: nitpick: card/power/relic descriptions generally end in a period

src/main/java/stsjorbsmod/powers/BurningPower.java
wang429: Hmm, BurningPower was displaying the number rounded up, and BurningLoseHpAction was using a rounded down value.
Some rewording would allow for some reusable code.

dbjorge: I think you could still do it with the current wording if you used the logic here in BurningPower for the calculation and had `BurningLoseHpAction` do `p.amount -= p.calculateNextFalloffAmount()`

wang429: I'm going to go with `calculateNextBurningAmount()` so I don't have to do endless subtractions.

Issue #151, opened 2019-11-14
Finding "Divine Fountain" event b/c of Grimoire


johnplunk:
Divine Fountain event is to remove all curses from deck, which is triggered by having >= 1 curses in deck (excluding Ascender's Bane). Grimoire proc's this event by being a curse (even though unremovable).

Following through with the "Remove all curses" event does not actually remove Grimoire though.

Frenoss:
Thanks for reporting this issue. Do you know if you encountered it on version 0.37.0 or an earlier version?

johnplunk:
Found it 0.37!

Frenoss:
Good deal. The next version should have a fix. Thanks!

dbjorge:
fixed by #157

PR #157, opened 2019-11-14
Finish support for Legendary cards

### LINKED ISSUE #131: "Legendary" card keyword/field to prevent duplication/transform/remove
*note: this is going to be a fairly complicated effect; if you're looking for a good first issue to tackle, this probably isn't it*

Per discord discussion, we want to add this for `ForbiddenGrimoire` and for several of the proposed Explorer cards later. Should act similarly to Ascender's Bane:

> Legendary cards cannot be duplicated, transformed, or removed from your deck.

Should still allow exhaust (different from necronomicurse)

Specific interactions should include (but probably won't be limited to):
* Patch `Astrolabe::onEquip` to ignore them
* Patch `CardGroup::getPurgeableCards` to omit them (but ideally also patch earlier rewards/etc to avoid choosing them as input)
* Patch `CardLibrary::getCurse` to omit them (if necessary? not sure if they exist to start)
* Add `AbstractCard.onRemoveFromMasterDeck` effect similar to `Necronomicurse`'s
* Patch `DollysMirror.onEquip` to omit it (this will probably need an `Instrument` patch)
* Patch `AbstractDungeon.transformCard` to omit them (but ideally also patch earlier rewards/etc to avoid choosing them as input) - hopefully the `getPurgableCards` patch covers this for most cases, eg it does for `NeowReward`'s transform option
    * Note, they shouldn't be candidates to be transformed to OR from
* string search through base game jar for other cases that special case IDs "AscendersBane" or "Necronomicurse"

Specific interactions to test include (not an exhaustive list):
* Card remove: `NewoReward`, bird cage, shop remove, the curse remove event, the pay health to remove, beggar, toke
* Transform: astrolabe, `NeowReward`, events
* Duplicate: `DollysMirror`

Frenoss:
Resolves #131 and #151.

Fix the Duplicator event. Now the player can't select a Legendary card to copy.
Fix the Fountain of Curse Removal event from happening if the player only has Legendary curses.
Log if we ever remove or transform a Legendary card that's in the master deck.
Support the Draft and SealedDeck mods. Comment our weirder interactions with other mods.

I think there are actual bugs in the main game for the Living Wall and Designer events. They check a condition different than what they use immediately afterwards places that are key to causing trouble for us. I don't plan to investigate how to file a bug with them, though.

### DIFF
--- a/src/main/java/stsjorbsmod/JorbsMod.java
+++ b/src/main/java/stsjorbsmod/JorbsMod.java
@@ -78,7 +78,11 @@ public class JorbsMod implements
         //   because this mod is not designed to interact with colored cards from the main game.
         //   Known cards: Dual Wield, Nightmare
         // - Living Wall event: if the player has no purgeable cards, they won't be given a chance to upgrade.
+        //     Spire bug #21944: Living Wall event checks for purgeable cards
+        //     in the Grow choice instead of upgradeable cards.
         // - Designer event: the Remove option can be active without any purgeable cards to select from.
+        //     Spire bug #21945: Designer event checks size of master deck for
+        //     enabling Remove/Transform option instead of purgeable cards.
         // - All Star mod won't give any Legendary colorless cards (should they ever exist).
         // - Hoarder mod will add two additional copies of Legendary cards.
         // - Insanity mod has access to Legendary cards and can generate duplicates.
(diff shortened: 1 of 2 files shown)


wang429:
I think just mentioning it in the official discord or reddit is good enough

dbjorge:
Specifically, if you go to the official slay the spire discord to the #bugs-and-feedback channel, you can say "bug: description of issue" and a bot will find the issue in their but tracker

Frenoss:
Thanks. I handed those two bugs to the bot and added comments for them in our code.

PR #160, opened 2019-11-16
Thirsting Sword: Card and Temp Art

evantor:
Added card Thirsting Sword. 
Added bad art for Thirsting Sword (Small …and Large). 
Modified card strings to include Thirsting Sword name and description.


src/main/java/stsjorbsmod/cards/wanderer/ThirstingSword.java
```diff
@@ -0,0 +1,81 @@
...
+
+        exhaust = true;
+
+        tags.add(CardTags.HEALING);
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        addToBot(new DamageAction(m, new DamageInfo(p, damage), AbstractGameAction.AttackEffect.FIRE)); //Deal Damage
+        addToBot(new ApplyPowerAction(m, p, new BurningPower(m, p, BURNING))); //Apply Burning
```
Frenoss: I think you mean `magicNumber` instead of `BURNING`

src/main/java/stsjorbsmod/cards/wanderer/ThirstingSword.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public void upgrade() {
+        if (!upgraded) {
+            upgradeName();
+            upgradeDamage(UPGRADE_DAMAGE);
+            upgradeMagicNumber(UPGRADE_BURNING);
+            upgradeMetaMagicNumber(UPGRADE_LOSE_MAX_HP);
+        }
```
Frenoss: We use upgradeDescription() at the end of all upgrade() to facilitate potential text changes either in the future or in localization.

src/main/java/stsjorbsmod/cards/wanderer/ThirstingSword.java
```diff
@@ -0,0 +1,82 @@
...
+                if (possibleExistingBurningPower != null) {
+                    int healAmount = possibleExistingBurningPower.amount;
+                    addToBot(new HealAction(p, p, healAmount)); //Heal for total burning on target
+                }
+                isDone = true;
+            }
+        });
+
+        int loseMaxHPAmount = -metaMagicNumber;
+        AbstractDungeon.player.increaseMaxHp(loseMaxHPAmount, true); //Lose MAX HP
```
dbjorge: Why not `decreaseMaxHealth`?

src/main/java/stsjorbsmod/cards/wanderer/ThirstingSword.java
```diff
@@ -0,0 +1,82 @@
...
+        exhaust = true;
+
+        tags.add(CardTags.HEALING);
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        addToBot(new DamageAction(m, new DamageInfo(p, damage), AbstractGameAction.AttackEffect.FIRE)); //Deal Damage
+        addToBot(new ApplyPowerAction(m, p, new BurningPower(m, p, magicNumber))); //Apply Burning
+        addToBot(new AbstractGameAction() {
```
dbjorge: I think this would have an issue where it would be removed by `GameActionManager.clearPostCombatActions` before the heal got queued if the `DamageAction` were to kill the last enemy. We discussed this a bit in the discord and agreed that reordering the card effects such that it went "burning -> heal ->-maxHp -> damage" would be a reasonable way to solve the ordering issue.

I think to implement that, you'd want to:

* move the `DamageAction` down to queue it after this custom action
* use `addToTop` instead of `addToBot` to queue the `HealAction`
* move the `decreaseMaxHealth` call into the custom action`

evantor: Thanks for the feedback!  I updated the card accordingly.  My one concern is that it doesn't synergize as well with the Lust memory anymore as now the burn gained from the damage dealt at the end doesn't count towards the heal.  But that might be minor enough to not matter.  If so, then the card should be good to go.

Issue #164, opened 2019-11-17
Purge Mind


wang429:
2: Deal 8 damage for all cards which remember in deck (including exhaust pile). Exhaust all cards which remember. Snap.

wang429:
TODO:
- Check each card for "Remember" effect and add it to the card's card
- Do some mspaint word art
- See how bad the game feels with slow exhaust effect and 5/10/15/20 remember cards
- Play around with VFX and SFX

PR #180, opened 2019-11-18
Purge mind

wang429:
addresses #164 

Purge Mind card
Add REMEMBER_MEMORY tag to a few "remember" cards

Assumptions:
- Similar Fiend Fire, exhausts cards then multi-attacks
- "cards in deck" means "cards in draw, hand, discard, and exhaust piles" rather than base deck.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -393,5 +393,9 @@
   "stsjorbsmod:FlameWard": {
     "NAME": "Flame Ward",
     "DESCRIPTION": "The next time you are attacked, apply !M! stsjorbsmod:Burning to ALL attackers and gain !B! Block. Exhaust."
+  },
+  "stsjorbsmod:PurgeMind": {
+    "NAME": "Purge Mind",
+    "DESCRIPTION": "Exhaust all cards which \"remember\". NL Deal !stsjorbsmod:BaseDamage! for each card Exhausted. Exhaust. NL Snap."
```
dbjorge: I think this description is a little off from the effect; it deals the damage even for cards which were already exhausted.

dbjorge: I think this should use `!D!` rather than `!stsjorbsmod:BaseDamage!`, since the damage-per-card is going to be modified by strength/weak/etc

dbjorge: `Snap` is a keyword, should be `stsjorbsmod:Snap`

wang429: Twas a relic of an initial implementation, can change.

dbjorge: Verified on discord that it's okay to reorder it from the spreadsheet description as "exhaust then damage" rather than "damage then exhaust" if that helps make the card text shorter, though

Issue #189, opened 2019-11-19
Update Amnesia


wang429:
New effect: At end of turn lose one random buff.

PR #190, opened 2019-11-20
Amnesia and Mage Armor

wang429:
Revise Amnesia #189 
Fix Mage Armor+ damage prevention amount


src/main/java/stsjorbsmod/cards/wanderer/Amnesia.java
```diff
@@ -38,9 +41,19 @@ public Amnesia() {
...
+                    .filter(power -> power.type == AbstractPower.PowerType.BUFF)
+                    .collect(Collectors.toList());
+            if (buffs.size() == 0) {
+                AbstractDungeon.actionManager.addToBottom(new TalkAction(true,"~...~", 2.0f, 2.0f));
+            } else if (p.hasPower(ArtifactPower.POWER_ID)) {
+                AbstractDungeon.actionManager.addToBottom(new ReducePowerAction(p, p, ArtifactPower.POWER_ID, 1));
+            } else {
+                int index = AbstractDungeon.miscRng.random(buffs.size());
+                AbstractPower powerToRemove = buffs.get(index);
+                AbstractDungeon.actionManager.addToBottom(new ReducePowerAction(p, p, powerToRemove.ID, powerToRemove.amount));
```
dbjorge: @Bobertg#0304 Is there some specific reason to prefer `new ReducePowerAction(p, p, powerToRemove.ID, powerToRemove.amount)` vs `new RemoveSpecificPowerAction(p, p, powerToRemove.ID)`? The latter feels more readable if there's no intended mechanical difference

src/main/java/stsjorbsmod/cards/wanderer/Amnesia.java
```diff
@@ -38,9 +41,19 @@ public Amnesia() {
...
+        if (dontTriggerOnUseCard) {
+            List<AbstractPower> buffs = p.powers.stream()
+                    .filter(power -> power.type == AbstractPower.PowerType.BUFF)
+                    .collect(Collectors.toList());
+            if (buffs.size() == 0) {
+                AbstractDungeon.actionManager.addToBottom(new TalkAction(true,"~...~", 2.0f, 2.0f));
+            } else if (p.hasPower(ArtifactPower.POWER_ID)) {
+                AbstractDungeon.actionManager.addToBottom(new ReducePowerAction(p, p, ArtifactPower.POWER_ID, 1));
+            } else {
+                int index = AbstractDungeon.miscRng.random(buffs.size());
```
dbjorge: The consensus from the main Slay the Spire #modding channel was that random card effects like this should generally be using `AbstractDungeon.cardRandomRng`

dbjorge: I think the argument to random is off-by-one; the range of `random(n)` is 0 through n *inclusive* (they did something different from most other random APIs I've worked with)

src/main/java/stsjorbsmod/cards/wanderer/Amnesia.java
```diff
@@ -38,9 +41,19 @@ public Amnesia() {
...
-            addToBot(new SnapAction(p));
-            addToBot(new RemovePowersMatchingPredicateAction(p, power -> power.type == AbstractPower.PowerType.BUFF || power.type == AbstractPower.PowerType.DEBUFF));
+        if (dontTriggerOnUseCard) {
+            List<AbstractPower> buffs = p.powers.stream()
+                    .filter(power -> power.type == AbstractPower.PowerType.BUFF)
+                    .collect(Collectors.toList());
+            if (buffs.size() == 0) {
+                AbstractDungeon.actionManager.addToBottom(new TalkAction(true,"~...~", 2.0f, 2.0f));
+            } else if (p.hasPower(ArtifactPower.POWER_ID)) {
+                AbstractDungeon.actionManager.addToBottom(new ReducePowerAction(p, p, ArtifactPower.POWER_ID, 1));
```
dbjorge: It might be more consistent looking for this to include the other effects that happen when Artifact triggers during power application (see `ApplyPowerAction`), something like:

```suggestion
                    addToBot(new TextAboveCreatureAction(this.target, EXTENDED_DESCRIPTION[0]));
                    CardCrawlGame.sound.play("NULLIFY_SFX");
                    p.getPower(ArtifactPower.POWER_ID).flashWithoutSound();
                    p.getPower(ArtifactPower.POWER_ID).onSpecificTrigger();
```

src/main/java/stsjorbsmod/cards/wanderer/Amnesia.java
wang429: Yeah, I remember changing what I was doing many times and forgot to bring this back to n-1

src/main/java/stsjorbsmod/cards/wanderer/Amnesia.java
wang429: Makes sense. I'll add this.

src/main/java/stsjorbsmod/cards/wanderer/Amnesia.java
wang429: I remember having issues with RemoveSpecificPowerAction in the past, which is why I didn't do it here. But maybe I was doing something wrong then.

src/main/java/stsjorbsmod/cards/wanderer/Amnesia.java
wang429: what is miscRng meant to be used for?

PR #206, opened 2019-11-27
Fragile mind nloths gift

wang429:
1. Fragile Mind can't be offered to N'loth. Tested by entering the event with any other relics (total 3) multiple times using `event N'loth`.
2. N'loth event can't occur if there are only two relics, Fragile Mind and any other. Unsure how to test.

### DIFF
--- a/src/main/java/stsjorbsmod/patches/NlothsGiftPatch.java
+++ b/src/main/java/stsjorbsmod/patches/NlothsGiftPatch.java
@@ -61,20 +61,4 @@ public class NlothsGiftPatch {
             return new ClonePlayerRelicsWithoutFragileMind();
         }
     }
-
-    @SpirePatch(clz = AbstractDungeon.class, method = "getShrine", paramtypez = { Random.class })
-    public static class AbstractDungeon_getShrine_NlothLogger {
-        @SpireInsertPatch(locator = NlothsGiftPatch.AbstractDungeon_getShrine_NlothLogger.Locator.class, localvars = "tmp")
-        public static void patch(Random rng, ArrayList<String> tmp) {
-            JorbsMod.logger.info(tmp);
-        }
-
-        public static class Locator extends SpireInsertLocator {
-            @Override
-            public int[] Locate(CtBehavior ctBehavior) throws Exception {
-                Matcher matcher = new Matcher.MethodCallMatcher(ArrayList.class, "get");
-                return LineFinder.findInOrder(ctBehavior, matcher);
-            }
-        }
-    }
 }
--- a/src/main/java/stsjorbsmod/patches/NlothsGiftPatch.java
+++ b/src/main/java/stsjorbsmod/patches/NlothsGiftPatch.java
@@ -1,13 +1,18 @@
 package stsjorbsmod.patches;
 
-import com.evacipated.cardcrawl.modthespire.lib.SpirePatch;
+import com.evacipated.cardcrawl.modthespire.lib.*;
 import com.megacrit.cardcrawl.characters.AbstractPlayer;
 import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.events.AbstractEvent;
 import com.megacrit.cardcrawl.events.shrines.Nloth;
+import com.megacrit.cardcrawl.random.Random;
 import com.megacrit.cardcrawl.relics.AbstractRelic;
 import javassist.CannotCompileException;
+import javassist.CtBehavior;
 import javassist.expr.ExprEditor;
 import javassist.expr.FieldAccess;
+import javassist.expr.MethodCall;
+import stsjorbsmod.JorbsMod;
 import stsjorbsmod.relics.FragileMindRelic;
 
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


src/main/java/stsjorbsmod/patches/NlothsGiftPatch.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @SpirePatch(clz = AbstractDungeon.class, method = "getShrine")
+    public static class AbstractDungeon_getShrine_NlothCheck {
+        public static ExprEditor Instrument() {
+            return new ExprEditor() {
+                @Override
+                public void edit(FieldAccess fieldAccess) throws CannotCompileException {
+                    if (fieldAccess.getClassName().equals(AbstractPlayer.class.getName())
+                            && fieldAccess.getFieldName().equals("relic")) {
```
dbjorge: I think this should be `relics` plural

src/main/java/stsjorbsmod/patches/NlothsGiftPatch.java
```diff
@@ -0,0 +1,81 @@
...
+        removeFragileMindRelic(relics.iterator());
+        return relics;
+    }
+
+    @SpirePatch(clz = AbstractDungeon.class, method = "getShrine")
+    public static class AbstractDungeon_getShrine_NlothCheck {
+        public static ExprEditor Instrument() {
+            return new ExprEditor() {
+                @Override
+                public void edit(FieldAccess fieldAccess) throws CannotCompileException {
```
dbjorge: For testing it, I think a good option would would be to add a second temporary path to getShrine that uses `JorbsMod.logger` to print out the `tmp` list after it's generated.

src/main/java/stsjorbsmod/patches/NlothsGiftPatch.java
```diff
@@ -0,0 +1,81 @@
...
+     * @param playerRelics
+     * @return
+     */
+    public static List<AbstractRelic> clonePlayerRelicsWithoutFragileMind(List<AbstractRelic> playerRelics) {
+        List<AbstractRelic> relics = new ArrayList<>(playerRelics);
+        removeFragileMindRelic(relics.iterator());
+        return relics;
+    }
+
+    @SpirePatch(clz = AbstractDungeon.class, method = "getShrine")
```
dbjorge: I think you could actually reuse exactly this same patch on `Nloth.class`/`SpirePatch.CONSTRUCTOR`, if you wanted to gain confidence that the patch implementation worked in the case that's easier to test. You can apply multiple `@SpirePatch` annotations to the same class.

src/main/java/stsjorbsmod/patches/NlothsGiftPatch.java
```diff
@@ -0,0 +1,81 @@
...
+    public static class AbstractDungeon_getShrine_NlothCheck {
+        public static ExprEditor Instrument() {
+            return new ExprEditor() {
+                @Override
+                public void edit(FieldAccess fieldAccess) throws CannotCompileException {
+                    if (fieldAccess.getClassName().equals(AbstractPlayer.class.getName())
+                            && fieldAccess.getFieldName().equals("relic")) {
+                        fieldAccess.replace("{ $_ = " + NlothsGiftPatch.class.getName() + ".clonePlayerRelicsWithoutFragileMind($0)); }");
```
dbjorge: `$0` isn't what you want here; when you're doing a field access patch for original code `someObject.someField`, `$0` represents `someObject`. What you're going to want to do is going to look almost identical to the `CloneMasterDeckWithoutLegendaryCardsEditor` class in `LegendaryPatch`; I'd recommend taking a look at what that's doing :)

src/main/java/stsjorbsmod/patches/NlothsGiftPatch.java
dbjorge: (or alternately, have separate @SpirePatch classes but have them both be based on a shared `ExprEditor` class, like how `LegendaryPatch.java` reuses `CloneMasterDeckWithoutLegendearyCardsEditor` for a few different `@SpirePatch`es)

src/main/java/stsjorbsmod/patches/NlothsGiftPatch.java
wang429: facepalming intensifies.

src/main/java/stsjorbsmod/patches/NlothsGiftPatch.java
wang429: Hmm I'll try this.

src/main/java/stsjorbsmod/patches/NlothsGiftPatch.java
wang429:
```
02:07:32.293 INFO stsjorbsmod.JorbsMod> [Match and Keep!, Wheel of Change, Golden Shrine, Transmorgrifier, Upgrade Shrine, Accursed Blacksmith, Designer, Duplicator, FaceTrader, Fountain of Cleansing, Lab, NoteForYourself, The Joust, The Woman in Blue]
```
output from that line with only the starting relics. N'loth is not there.

PR #219, opened 2019-12-05
Dynamic card text update

sanity-dance:
Add dynamic card descriptions to Animate Objects, Mania, and Magic Missiles.

### DIFF
--- a/src/main/java/stsjorbsmod/cards/wanderer/Mania.java
+++ b/src/main/java/stsjorbsmod/cards/wanderer/Mania.java
@@ -51,6 +51,42 @@ public class Mania extends CustomJorbsModCard {
     }
 
     @Override
+    public void applyPowers() {
+        int count = calculateBonusBaseDamage();
+        if(count>0) {
+            super.applyPowers();
+            if(this.upgraded) {
+                this.rawDescription = cardStrings.UPGRADE_DESCRIPTION + cardStrings.EXTENDED_DESCRIPTION[0];
+            }
+            else
+                this.rawDescription = cardStrings.DESCRIPTION + cardStrings.EXTENDED_DESCRIPTION[0];
+            initializeDescription();
+        }
+    }
+
+    @Override
+    public void onMoveToDiscardImpl() {
+        if(this.upgraded) {
+            this.rawDescription = cardStrings.UPGRADE_DESCRIPTION;
+        }
+        else
+            this.rawDescription = cardStrings.DESCRIPTION;
+        initializeDescription();
+    }
+
+    @Override
+    public void calculateCardDamage(AbstractMonster mo) {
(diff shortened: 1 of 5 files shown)


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -15,7 +15,10 @@
   },
   "stsjorbsmod:AnimateObjects": {
     "NAME": "Animate Objects",
-    "DESCRIPTION": "Deal !D! damage to ALL enemies for each non-exhausted card that didn't start the combat in your deck."
+    "DESCRIPTION": "Deal !D! damage for each non-exhausted card added to your deck this combat to ALL enemies.",
```
dbjorge: iryan72 in Discord suggested that they thought `"Deal !D! damage to ALL enemies for each non-exhausted card added to your deck this combat."` read a little more naturally; I agree with that feedback

src/main/java/stsjorbsmod/cards/wanderer/Mania.java
```diff
@@ -50,6 +50,42 @@ private boolean isEligibleForExtraEffect() {
...
+        }
+        else
+            this.rawDescription = cardStrings.DESCRIPTION;
+        initializeDescription();
+    }
+
+    @Override
+    public void calculateCardDamage(AbstractMonster mo) {
+        super.calculateCardDamage(mo);
+        if(this.upgraded) {
```
dbjorge: For the sake of localization (where there may be cases where some languages require UPGRADE_DESCRIPTIONs even when english does not because of different pluralization rules/etc), I think it would be a good idea for all of the different cards with this sort of logic to include this "if upgraded" check even if it isn't necessary in english for all the cards.

Maybe using helpers in `CustomJorbsModCard` would help reduce the boilerplate for doing this? Something like

```java
// CustomJorbsModCard
protected String rawDynamicDescriptionSuffix = "";

@Override
public void initializeDescription() {
    String originalRawDescription = this.rawDescription;
    this.rawDescription += this.rawDynamicDescriptionSuffix;
    super.initializeDescription();
    this.rawDescription = originalRawDescription;
}

// Individual cards...
@Override
public void applyPowers() {
    super.applyPowers();
    this.rawDynamicDescriptionSuffix = cardStrings.EXTENDED_DESCRIPTION[0];
}
@Override
public void calculateCardDamage(AbstractMonster mo) {
    super.calculateCardDamage(mo);
    this.rawDynamicDescriptionSuffix = cardStrings.EXTENDED_DESCRIPTION[0];
}
@Override
public void onMoveToDiscardImpl() {
    this.rawDynamicDescriptionSuffix = "";
}
```

PR #220, opened 2019-12-06
Patron

wang429:
The downgrade portion doesn't quite work. Playing Patron+ still seems to appear as Patron in the exhaust pile. The master deck has the downgraded version. It's probably just a matter of checking exhaust pile (or discard pile with spoon relic) for the card.

Downgrade assumption being made is that we are only going to downgrade specific mod cards and only going to downgrade upgraded cards (IE no Patron-)


src/main/java/stsjorbsmod/actions/MemoryManagerSnapAction.java
```diff
@@ -0,0 +1,18 @@
+package stsjorbsmod.actions;
+
+import com.megacrit.cardcrawl.actions.AbstractGameAction;
+import stsjorbsmod.memories.MemoryManager;
+
+public class MemoryManagerSnapAction extends AbstractGameAction {
```
dbjorge: Was this intended to be part of this PR? (is it necessary at all? normally you wouldn't want to invoke snap outside of the fuller SnapAction that already exists)

src/main/java/stsjorbsmod/cards/wanderer/Patron.java
```diff
@@ -0,0 +1,84 @@
...
+        addToBot(new DamageAction(m, new DamageInfo(p, damage)));
+
+        Optional<AbstractCard> cardOp = AbstractDungeon.player.masterDeck.group.stream().filter(c -> c.uuid == uuid).findAny();
+        if(!cardOp.isPresent()) {
+            JorbsMod.logger.info("Failed to purge a card we didn't have. Perhaps Duplication Potion or Attack Potion or similar effect occurred.");
+            return;
+        }
+        // In the highly unlikely event that UUIDs clash and the found card isn't actually the specific Patron card... oops
+        AbstractCard masterCard = cardOp.get();
+        if (upgraded) {
```
dbjorge: For both downgrade and upgrade effects, I'd recommend passing the actions UUIDs rather than cards instances and using those to search for all in-battle instances, as well (this should resolve the issue you mention in the PR description):

```java
masterCopy = StSLib.getMasterDeckEquivalent(originalCard)
allInstances = GetAllInBattleInstances.get(originalCard.uuid);
if (masterCopy != null) {
    allInstances.add(masterCopy);
}
for (AbstractCard c : allInstances) {
    // perform purge / downgrade effect
}
```

src/main/java/stsjorbsmod/actions/DestroyCardAction.java
```diff
@@ -0,0 +1,30 @@
...
+        this.card = card;
+    }
+
+    @Override
+    public void update() {
+        JorbsMod.logger.info("DestroyCardAction purging " + card.toString());
+        CardCrawlGame.sound.play("CARD_EXHAUST");
+        AbstractDungeon.topLevelEffects.add(new PurgeCardEffect(card, (float)(Settings.WIDTH / 2), (float)(Settings.HEIGHT / 2)));
+        AbstractDungeon.player.masterDeck.removeCard(card);
+        isDone = true;
```
dbjorge: This should probably have a "fast" duration similar to `UpgradeSpecificCardAction` and `ExhaustSpecificCardAction` to give the animation time to happen

src/main/java/stsjorbsmod/actions/DowngradeCardAction.java
```diff
@@ -0,0 +1,23 @@
...
+    private DowngradeableCard card;
+
+    public DowngradeCardAction(DowngradeableCard card) {
+        this.card = card;
+    }
+
+    @Override
+    public void update() {
+        JorbsMod.logger.info("DowngradeCardAction downgrading " + card.toString());
+        card.downgrade();
```
dbjorge: It would be nice for this to have an sfx/vfx (reusing an existing one for now would be fine to start). If you do this, same comment as on DestroyCardAction about giving it a duration

src/main/java/stsjorbsmod/actions/MemoryManagerSnapAction.java
wang429: nope, I thought I had backed stuff out that i was playing around with. but clearly not.

src/main/java/stsjorbsmod/cards/wanderer/Patron.java
wang429: oh didn't realized that was a thing. seems good. though I probably don't need the allInstances anymore.

wang429:
Latest push doesn't address comments, but I think does allow downgrade and destroy to work as far as I can tell.
There is an issue where base Patron doesn't get duplicated with duplication potion. duplication effects check for `!card.purgeOnUse`

dbjorge:
> There is an issue where base Patron doesn't get duplicated with duplication potion. duplication effects check for !card.purgeOnUse

I think if you were to update the destroy instance to also cover the in battle instances, you could work around this by omitting `purgeOnUse` (I think it's generally intended more for temporary cards like the ones generated for Echo Form)

wang429:
Things should be functional now.
Destroy effect is faster
Downgrade effect is just display the downgraded card. Probably will adjust this later.
May need to check if playing the card counts as exhausting it. Strange Spoon does strange things. Actually Strange Spoon does strange things with ephemeral. Card is actually discarded, then pops into exhaust pile.

PR #227, opened 2019-12-09
Grade change fx

wang429:
Wrath upgrade: simple VFX and SFX
Downgrade: simple VFX


wang429:
Current issue with implementation: when Patron+ kills a creature, both wrath and downgrade seem to be rendered. Need to check with Patron and destroy + wrath

wang429:
Playing Patron with wrath doesn't bother with wrath stacks anymore.
Playing Patron or Patron+ now destroys or downgrades properly.
Playing Patron+ still stacks downgrade and wrath a little bit, but since effects happen at the same time, I'm not sure exactly how to handle that.

src/main/java/stsjorbsmod/cards/wanderer/Patron.java
```diff
@@ -61,7 +54,7 @@ public void downgrade() {
             upgraded = false;
             name = originalName;
             initializeTitle();
-            baseDamage = DAMAGE;
+            baseDamage = DAMAGE + WrathField.wrathEffectCount.get(this);
```
dbjorge: There might be other mod effects for this; probably the most reliable option would be to stick as close to how upgrading works as possible (ie, subtract `UPGRADE_PLUS_DAMAGE` rather than resetting)

src/main/java/stsjorbsmod/actions/PatronAction.java
```diff
@@ -0,0 +1,59 @@
...
+            }
+            if (AbstractDungeon.getCurrRoom().monsters.areMonstersBasicallyDead()) {
+                AbstractDungeon.actionManager.clearPostCombatActions();
+            }
+        }
+
+        tickDuration();
+        if (isDone) {
+            if (isUpgraded) {
+                showDowngradeEffect(card, duration);
```
dbjorge: Is `duration` actually the field you mean to use here? It's the one `tickDuration` is decrementing, so it's always going to be negative at this point if `tickDuration` has set `isDone = true`

src/main/java/stsjorbsmod/memories/WrathMemory.java
```diff
@@ -94,6 +94,12 @@ private void permanentlyIncreaseCardDamage(AbstractCard card) {
             return;
         }
 
+        // ignore cards that are being removed from the deck
```
dbjorge: I don't think I agree with this change; I understand it's mechanically not different, but I think showing the vfx is important so that the player understands what happened. If anything, I think it might make sense to addToTop a WaitAction to go with the wrath effect, to give the player time to process it (that might also help make the Patron+ effect chain more understandable)

src/main/java/stsjorbsmod/cards/wanderer/Patron.java
```diff
@@ -36,13 +35,7 @@ public Patron() {
...
     @Override
     public void use(AbstractPlayer p, AbstractMonster m) {
-        addToBot(new DamageAction(m, new DamageInfo(p, damage)));
-
-        if (upgraded) {
-            addToBot(new DowngradeCardPermanentlyAction(this));
-        } else {
-            addToBot(new DestroyCardAction(this));
-        }
+        addToBot(new PatronAction(m, new DamageInfo(p, damage), this));
```
dbjorge: I think I actually preferred the model where it was separate actions for damage vs the downgrade/destroy; I think it would make more sense for downgrade/destroy to own their own respective effect timings.

wang429: It was necessary to do this because, like with Wrath, the downgrade/destroy should still occur if the card ended the combat. This was done following Watcher's upgrade card `LessonLearned`

wang429: or were you thinking maybe something like PatronDowngradeAction and PatronDestroyAction both do damage, but do their respective downgrade/destroy?

src/main/java/stsjorbsmod/actions/PatronAction.java
wang429: This duration isn't even necessary.

src/main/java/stsjorbsmod/cards/wanderer/Patron.java
wang429: Looking over it again, the effect timings aren't actually tied to this action.

Issue #230, opened 2019-12-09
20191208-20191209 balance changes


wang429:
> balance pass from tonight:
> - [ ] burning 50% falloff -> 33% falloff
> - [ ] burning prevents all healing
>
> functionality changes:
> - [ ] make burning prevent heal on time eater/champ even though they cleanse
> - [x] make cards which exhume when snap at end of turn 7 go into starting hand for turn 8

wang429:
Implementation notes, ideally move snap after turn end card discard. If that can't work, maybe have fragile mind's snap trigger exhume at start of enemy turn?

wang429:
The exhume on snap is its own ticket #228

dbjorge:
Burning updates resolved by #233, rest moved to separate issue

PR #233, opened 2019-12-11
1209 balance changes

wang429:
BUFF: burning - amount reduced by down from 50% to 33% (1/2 to 1/3)
BUFF: burning - healing prevented (before, reduced by 50%)
BUFF: burning - Time Eater's healing is now prevented if it was on fire

addresses #230


src/main/java/stsjorbsmod/patches/BurningTimeEaterPatch.java
```diff
@@ -0,0 +1,56 @@
...
+import stsjorbsmod.powers.BurningPower;
+
+import java.util.Iterator;
+
+public class BurningTimeEaterPatch {
+    private static final String UI_ID = JorbsMod.makeID(BurningTimeEaterPatch.class);
+    private static final String[] TEXT = CardCrawlGame.languagePack.getUIString(UI_ID).TEXT;
+
+    @SpirePatch(clz = RemoveDebuffsAction.class, method = "update")
+    public static class RemoveDebuffsAction_Update_BurningTimeEater {
```
dbjorge: I think you could probably simplify and avoid a patch for this by instead adding something like this to `BurningPower`:

```java
    @Override
    public void onRemove() {
        // This makes the heal prevention affect Time Eater's "cleanse and heal simultaneously" turn
        // despite the cleanse technically happening first.
        for (AbstractGameAction a : AbstractDungeon.actionManager.actions) {
            if (TimeEater.ID.equals(this.owner.id) && a instanceof HealAction && a.target == this.owner) {
                a.amount = this.onHeal(a.amount);
            }
        }
    }
```

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Keyword-Strings.json
```diff
@@ -1,7 +1,7 @@
 [
   {
     "NAMES": ["burning"],
-    "DESCRIPTION": "Burning creatures take damage at the start of their turn and have healing reduced by 50%. Each turn, Burning is reduced by half."
+    "DESCRIPTION": "Burning creatures take damage at the start of their turn and have healing prevented. Each turn, Burning is reduced by a third."
```
dbjorge: I feel like "have healing prevented" sounds a little ambiguous about whether it prevents them from healing others; maybe "cannot be healed" would be more clear, more like how you've update the power strings?

src/main/java/stsjorbsmod/patches/BurningTimeEaterPatch.java
wang429: Oh yeah... forgot to check the overrideable methods on AbstractPower... Yeah I'll take a look and see.
Another thing I was considering was to simply remove the healing action all together. Heal for 0 doesn't have the animation as far as I can tell.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Keyword-Strings.json
wang429: Sure, that makes sense.

PR #250, opened 2019-12-20
Apparate

wang429:
Apparate implemented with a guess on its upgrade.
this allows the player to successfully win the first fight, and then immediately crash. We need some cards that aren't consider "BASIC" before card rewards work.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -20,6 +20,10 @@
       " NL (Gain !M! Thorns.)"
     ]
   },
+  "stsjorbsmod:Apparate": {
+    "NAME": "Apparate",
+    "DESCRIPTION": "Gain !stsjorbsmod:UrMagic! Vulnerable. Deal !D! damage and apply !M! vulnerable to all enemies."
```
dbjorge: Capitalize second `vulnerable` in description line?

dbjorge: Usually the game uses all-caps `ALL` to refer to `ALL enemies`

src/main/java/stsjorbsmod/cards/cull/Apparate.java
```diff
@@ -0,0 +1,56 @@
...
+        baseDamage = DAMAGE;
+        magicNumber = baseMagicNumber = ENEMY_VULNERABLE;
+        urMagicNumber = baseUrMagicNumber = SELF_VULNERABLE;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster abstractMonster) {
+        addToBot(new ApplyPowerAction(p, p, new VulnerablePower(p, urMagicNumber, false)));
+        addToBot(new VFXAction(new CleaveEffect()));
+        addToBot(new DamageAllEnemiesAction(p, DamageInfo.createDamageMatrix(damage), DamageInfo.DamageType.NORMAL, AbstractGameAction.AttackEffect.NONE));
```
dbjorge: `DamageInfo.createDamageMatrix` is the suggested way for powers to calculate multi-damage, but for cards it's instead suggested to set `isMultiDamage = true;` in the constructor and use the built-in `multiDamage` field as the second argument to the damage all action; it results in more correct interaction with certain combinations of vulnerable/etc on only some of the enemies. See `Fireball` for an example.

PR #253, opened 2019-12-21
Cull cards

wang429:
CULL card and Inhale.
Upgrades are simply guesses.


src/main/java/stsjorbsmod/powers/StrengthNextTurnPower.java
```diff
@@ -0,0 +1,53 @@
...
+    public static final String NAME = powerStrings.NAME;
+    public static final String[] DESCRIPTIONS = powerStrings.DESCRIPTIONS;
+
+    private static final Texture tex84 = TextureLoader.getTexture(makePowerPath("strength_next_turn_power84.png"));
+    private static final Texture tex32 = TextureLoader.getTexture(makePowerPath("strength_next_turn_power32.png"));
+
+    private int damageAbsorbedThisTurn = 0;
+
+    public StrengthNextTurnPower(AbstractCreature owner, int amount) {
+        super(StaticPowerInfo.Load(StrengthNextTurnPower.class));
```
dbjorge: `StaticPowerInfo` is intended to be loaded into a static field, rather than each time it's constructed; loading the info is a little expensive, so it avoids some frame jitter. I still need to update the other existing powers to use the intended strategy, but `CoilPower` is a good example of how it's intended to be used

src/main/java/stsjorbsmod/cards/cull/CULLCard.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public void applyPowers() {
+        baseBlock = misc;
+        super.applyPowers();
+        initializeDescription();
+    }
+
+    @Override
+    public void onDraw() {
+        addToBot(new ModifyDamageAction(uuid, -urMagicNumber));
```
dbjorge: I think ModifyDamageAction only normally affects the in combat instances, but the spreadsheet suggests the effect is intended to be permanent (this probably implies using the misc field like Ritual Dagger does, instead of/in addition to updating baseDamage like ModifyDamageAction does)

src/main/java/stsjorbsmod/cards/cull/CULLCard.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        this.addToBot(new RitualDaggerAction(m, new DamageInfo(p, damage, damageTypeForTurn), magicNumber, uuid));
+        this.addToBot(new RitualDaggerAction(m, new DamageInfo(p, damage, damageTypeForTurn), magicNumber, uuid));
+    }
+
+    @Override
+    public void applyPowers() {
+        baseBlock = misc;
```
dbjorge: Is this necessary? I see that Ritual Dagger does this, but I can't figure out why; I suspect it might be an unnecessary copy/paste from Genetic Algorithm...

src/main/java/stsjorbsmod/cards/cull/CULLCard.java
```diff
@@ -0,0 +1,67 @@
...
+import stsjorbsmod.cards.CustomJorbsModCard;
+import stsjorbsmod.cards.OnDrawCardSubscriber;
+import stsjorbsmod.characters.Cull;
+
+import static stsjorbsmod.JorbsMod.JorbsCardTags.LEGENDARY;
+
+public class CULLCard extends CustomJorbsModCard implements OnDrawCardSubscriber {
+    public static final String ID = JorbsMod.makeID(CULLCard.class);
+
+    private static final CardRarity RARITY = CardRarity.SPECIAL;
```
dbjorge: CardRarity.BASIC, I think

src/main/java/stsjorbsmod/cards/cull/CULLCard.java
wang429: You know, I could have sworn I read what ModifyDamageAction did several times to make sure of exactly what you're saying...

src/main/java/stsjorbsmod/cards/cull/CULLCard.java
wang429: I'll see what happens and remove if unneeded

PR #255, opened 2019-12-21
Cull cards pt2

wang429:
Haunt
Taunt
Toil
Cull card changes allow for save/load. Still need to figure out why its damage is decreasing by 2 on draw. Allows for the player to climb a little bit before simply freezing at the end of combat.


wang429:
Figured out why the CULL card is dropping by 2 on draw. TFW you implement try to perform a shortcut in implementation only to duplicate the implementation anyways.

Still need to figure out what would be causing a game freeze post combat.

wang429:
Oh right, we need at least 3 of each rarity to guarantee card rewards don't crash because they check for uniqueness

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -192,6 +192,10 @@
     "DESCRIPTION": "Draw !M! cards. NL This turn, your next attack is played !stsjorbsmod:MetaMagic! additional time.",
     "UPGRADE_DESCRIPTION": "Draw !M! cards. NL This turn, your next attack is played !stsjorbsmod:MetaMagic! additional times."
   },
+  "stsjorbsmod:Haunt": {
+    "NAME": "Haunt",
+    "DESCRIPTION": "Gain !stsjorbsmod:UrMagic! Intangible. NL Next turn, Draw !M! cards. NL Exhaust."
```
dbjorge: Probably the most consistent wording for the draw effect would be to match Predator, something like `Next turn, draw !M! additional cards.` (lowercase draw, added word "additional")

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -455,6 +463,10 @@
     "DESCRIPTION": "Remember stsjorbsmod:Kindness. NL Gain stsjorbsmod:Clarity of current memory.",
     "UPGRADE_DESCRIPTION": "Innate. NL Remember stsjorbsmod:Kindness. NL Gain stsjorbsmod:Clarity of current memory."
   },
+  "stsjorbsmod:Toil": {
+    "NAME": "Toil",
+    "DESCRIPTION": "Deal !D! damage. NL If Fatal, permanently upgrade a card. NL Exhaust."
```
dbjorge: "a random card"?

dbjorge: On fatal vs allowing minions:

> [5:24 PM] brosmike: Is it intentional that the wording included minion kills?
> [5:29 PM] jorbs: probably

src/main/java/stsjorbsmod/patches/LoadMiscValuePatch.java
```diff
@@ -0,0 +1,33 @@
...
+import javassist.CtBehavior;
+import stsjorbsmod.cards.CustomJorbsModCard;
+
+import java.util.Arrays;
+import java.util.Collections;
+
+public class LoadMiscValuePatch {
+    @SpirePatch(clz = CardLibrary.class, method = "getCopy", paramtypez = {String.class, int.class, int.class})
+    public static class CardLibrary_getCopyPatch {
+        @SpireInsertPatch(locator = CardID_StringEqualsLocator.class, localvars = "retVal")
```
dbjorge: I would recommend using a postfix patch for this instead; this insert patch is relying on them never introducing some new special case based on cardID that isn't related to misc, which seems a little fragile.

Postfix patches can't get local vars but they do get a copy of the original return value, which is all you need here:

```java
@SpirePostfixPatch
public static AbstractCard patch(AbstractCard __result, String key, int upgradeTime, int misc) {
    if (misc != 0 && __result instanceof CustomJorbsModCard) {
        ((CustomJorbsModCard) __result).applyLoadedMiscValue(misc);
    }
    return __result;
}
```

src/main/java/stsjorbsmod/cards/cull/Toil.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public Toil() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        baseDamage = DAMAGE;
+        exhaust = true;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        addToBot(new LessonLearnedAction(m, new DamageInfo(p, damage)));
```
dbjorge: This looks fine as currently written but would need to change given the desire to have it work with minions, too

src/main/java/stsjorbsmod/cards/cull/Taunt.java
```diff
@@ -0,0 +1,51 @@
...
+import stsjorbsmod.JorbsMod;
+import stsjorbsmod.cards.CustomJorbsModCard;
+import stsjorbsmod.characters.Cull;
+
+public class Taunt extends CustomJorbsModCard {
+    public static final String ID = JorbsMod.makeID(Taunt.class);
+
+    private static final CardRarity RARITY = CardRarity.UNCOMMON;
+    private static final CardTarget TARGET = CardTarget.ALL;
+    private static final CardType TYPE = CardType.ATTACK;
```
dbjorge: Should probably be a skill (remember to regenerate the card art)

src/main/java/stsjorbsmod/cards/cull/CULLCard.java
```diff
@@ -41,7 +44,7 @@ public CULLCard() {
 
     @Override
     public void use(AbstractPlayer p, AbstractMonster m) {
-        this.addToBot(new RitualDaggerAction(m, new DamageInfo(p, damage, damageTypeForTurn), magicNumber, uuid));
+        this.addToBot(new DamageAction(m, new DamageInfo(p, damage, damageTypeForTurn), AbstractGameAction.AttackEffect.SLASH_VERTICAL));
```
dbjorge: Why this change? Isn't it desirable for it to upgrade even if the first swing kills?

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
wang429: Yeah... ;~; can't use LessonLearnedAction

src/main/java/stsjorbsmod/actions/PermanentlyModifyDamageAction.java
```diff
@@ -28,12 +28,6 @@ public void update() {
                 c.isDamageModified = false;
             }
         }
-
-        for(var1 = GetAllInBattleInstances.get(uuid).iterator(); var1.hasNext(); c.baseDamage = c.misc) {
```
dbjorge: I don't think the original `ModifyDamageAction` does what we want here, since it doesn't decrement `misc` and so isn't going to cooperate with `RitualDaggerAction`. I'd recommend not extending from it at all, personally; it's just not saving that much code.

src/main/java/stsjorbsmod/cards/cull/CULLCard.java
wang429: Yeah, forgot about that. Though, with `RitualDaggerAction` done twice, if the fight doesn't end, Cull card damage increases twice. Working on a new Action for it.

src/main/java/stsjorbsmod/actions/PermanentlyModifyDamageAction.java
dbjorge: In the loop above this, would be nice to convert to a for each

dbjorge: This should probably clamp misc/baseDamage to a minimum of 0

src/main/java/stsjorbsmod/actions/CullCardAction.java
```diff
@@ -0,0 +1,62 @@
...
+import com.megacrit.cardcrawl.cards.DamageInfo;
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.helpers.GetAllInBattleInstances;
+import com.megacrit.cardcrawl.vfx.combat.FlashAtkImgEffect;
+
+import java.util.Iterator;
+import java.util.UUID;
+
+public class CullCardAction extends AbstractGameAction {
```
dbjorge: Did you mean to include this? It doesn't look like CULLCard is using it anymore

src/main/java/stsjorbsmod/cards/cull/CULLCard.java
dbjorge: We'd probably need a new action anyway since CULL is also supposed to upgrade on minion kills

src/main/java/stsjorbsmod/cards/cull/Toil.java
```diff
@@ -0,0 +1,47 @@
...
+    public static final CardColor COLOR = Cull.Enums.CULL_CARD_COLOR;
+
+    private static final int COST = 2;
+    private static final int DAMAGE = 15;
+    private static final int UPGRADE_PLUS_DAMAGE = 5;
+
+    public Toil() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        baseDamage = DAMAGE;
+        exhaust = true;
```
dbjorge: Should probably have a metaupgrade card tag

src/main/java/stsjorbsmod/actions/CullCardAction.java
wang429: It will be needed to upgrade on first swing at the end of combat, and upgrade only once if combat not finished.

src/main/java/stsjorbsmod/cards/cull/CULLCard.java
wang429: oh right.

src/main/java/stsjorbsmod/JorbsMod.java
```diff
@@ -104,6 +104,10 @@
         // Use on a card that remembers a memory, which is mechanic specific to the Wanderer character.
         @SpireEnum(name = "REMEMBER_MEMORY")
         public static AbstractCard.CardTags REMEMBER_MEMORY;
+
+        // Use on a card that performs metaupgrades, like Toil.
+        @SpireEnum(name = "METAUPGRADE")
```
dbjorge: `PERSISTENT_POSITIVE_EFFECT` is the one to re-use (it already has the associated patches to remove them from temporary pools like skill potion)

src/main/java/stsjorbsmod/actions/ToilAction.java
```diff
@@ -0,0 +1,61 @@
...
+    public ToilAction(AbstractCreature target, DamageInfo info) {
+        this.info = info;
+        setValues(target, info);
+        actionType = AbstractGameAction.ActionType.DAMAGE;
+        duration = Settings.ACTION_DUR_MED;
+    }
+
+    public void update() {
+        if (duration == Settings.ACTION_DUR_MED && target != null) {
+            AbstractDungeon.effectList.add(new FlashAtkImgEffect(target.hb.cX, target.hb.cY, AbstractGameAction.AttackEffect.NONE));
```
dbjorge: Is this intended to be using `AttackEffect.NONE`?

wang429: nope

PR #259, opened 2019-12-24
Cull card shake

Cartopol:
Implemented Cull common card: Shake
(0: Draw 2(3) cards if target is vulnerable. Apply 1 vulnerable. Exhaust.)


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -396,6 +396,10 @@
     "NAME": "Scorching Ray",
     "DESCRIPTION": "Remember stsjorbsmod:Lust. NL Apply !stsjorbsmod:MetaMagic! stsjorbsmod:Burning !M! times. This stsjorbsmod:Burning is affected by Strength."
   },
+  "stsjorbsmod:Shake": {
+    "NAME": "Shake",
+    "DESCRIPTION": "If enemy is Vulnerable, draw !M! cards. NL Apply 1 Vulnerable. NL Exhaust."
```
dbjorge: You can use placeholders for the vulnerable amount too, it just looks a little different because it's a custom placeholder babe. I think the syntax is `!stsjorbsmod:UrMagic!`

src/main/java/stsjorbsmod/cards/cull/Shake.java
```diff
@@ -0,0 +1,47 @@
...
+            addToBot(new DrawCardAction(magicNumber));
+        }
+        addToBot(new ApplyPowerAction(abstractMonster, p, new VulnerablePower(p, urMagicNumber, false)));
+    }
+
+    @Override
+    public void upgrade() {
+        if (!upgraded) {
+            upgradeName();
+            upgradeMagicNumber(UPGRADE_PLUS_CARDS);
```
dbjorge: It's good to call `upgradeDescription();` here as well; even if English doesn't require updating the description on upgrade, sometimes other languages do, so we generally like to just always include it

src/main/java/stsjorbsmod/cards/cull/Shake.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public Shake() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        magicNumber = baseMagicNumber = CARDS_DRAWN;
+        urMagicNumber = baseUrMagicNumber = VULNERABLE;
+        exhaust = true;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster abstractMonster) {
```
dbjorge: For cards with effects that are conditional like this, the beta branch introduced a new bit of UI where the cards glow gold instead of blue when the "extra" effect is available; it would be good to implement that here based on whether any enemies are currently vulnerable. See Counterspell for an example.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
dbjorge: It's good to do this even if it isn't changing on upgrade right now because it means any future balance changes to the card mechanics are less likely to require updates for all localization files

PR #271, opened 2019-12-27
Implemented Cull's Consume Card mechanic + Frostbite card

Cartopol:
Includes:
- Generic ConsumeCardAction
- ConsumeCardIfFatalAction (required for Frostbite)
- Basic card: Frostbite (Deal 6(9) damage. If Fatal, consume this.)
- Frostbite added to Cull starter deck
- New keyword: Consume


src/main/java/stsjorbsmod/actions/ConsumeCardAction.java
```diff
@@ -0,0 +1,65 @@
...
+
+            case BASIC:
+                AbstractDungeon.player.heal(BASIC_HEAL);
+                break;
+            case COMMON:
+                AbstractDungeon.player.increaseMaxHp(COMMON_HP_GAIN, false);
+                break;
+
+            case SPECIAL:
+                if (card.hasTag(JorbsMod.JorbsCardTags.LEGENDARY)) {
```
dbjorge: LEGENDARY cards are not necessarily SPECIAL rarity; the tag check here should probably happen before/outside of the switch statement.

src/main/java/stsjorbsmod/actions/ConsumeCardAction.java
```diff
@@ -0,0 +1,65 @@
...
+    private static final int RARE_HP_GAIN = 5;
+    private static final int LEGENDARY_HP_GAIN = 5;
+
+    public ConsumeCardAction(AbstractCard c) {
+        this.card = c;
+    }
+
+    public void update() {
+        switch (card.rarity) {
+            case CURSE:
```
dbjorge: CURSE is also intended to decrease max hp by 1

src/main/java/stsjorbsmod/actions/ConsumeCardAction.java
```diff
@@ -0,0 +1,65 @@
...
+                AbstractDungeon.player.heal(AbstractDungeon.player.maxHealth);
+                break;
+
+            case RARE:
+                AbstractDungeon.player.increaseMaxHp(RARE_HP_GAIN, false);
+                AbstractDungeon.player.heal(AbstractDungeon.player.maxHealth);
+                break;
+
+        }
+        AbstractDungeon.actionManager.addToBottom(new ConsumerGameAction<>(CardMetaUtils::removeCard, card));
```
dbjorge: Do these need to be separate actions, given that the consume effect is already separated to an action?

src/main/java/stsjorbsmod/actions/ConsumeCardIfFatalAction.java
```diff
@@ -0,0 +1,35 @@
+//
+// Source code recreated from a .class file by IntelliJ IDEA
```
dbjorge: Can drop the comment

src/main/java/stsjorbsmod/actions/ConsumeCardIfFatalAction.java
```diff
@@ -0,0 +1,35 @@
...
+        this.info = info;
+        this.setValues(target, info);
+        this.card = c;
+        this.actionType = ActionType.DAMAGE;
+        this.duration = 0.1F;
+    }
+
+    public void update() {
+        if (this.duration == 0.1F && this.target != null) {
+            this.target.damage(this.info);
```
dbjorge: In general, any action that's directly invoking `monster.damage` needs a `clearPostCombatActions` check. I think maybe it would be helpful to make a generalized form of the action you implemented for Siphon, something like `DamageWithOnKillEffectAction` that accepts a Runnable; I think there's something like a half dozen CULL cards that could end up using that.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -195,6 +195,10 @@
     "NAME": "Fresh Adventure",
     "DESCRIPTION": "Deal !D! damage. NL Remember stsjorbsmod:Diligence."
   },
+  "stsjorbsmod:Frostbite": {
+    "NAME": "Frostbite",
+    "DESCRIPTION": "Deal !D! damage. NL If Fatal, stsjorbsmod:Consume this."
```
dbjorge: The sheet's description here is inclusive of minion kills; I like the idea you did in Siphon where you made "works on minions" just an argument to the action, so we can change this easily later.

src/main/java/stsjorbsmod/actions/ConsumeCardAction.java
dbjorge: `CardMetaUtils::removeCard` was renamed in a recent refactor, there will be a silent merge conflict here

src/main/java/stsjorbsmod/cards/cull/Frostbite.java
```diff
@@ -0,0 +1,41 @@
...
+    private static final CardType TYPE = CardType.ATTACK;
+    public static final CardColor COLOR = Cull.Enums.CULL_CARD_COLOR;
+
+    private static final int COST = 1;
+    private static final int DAMAGE = 6;
+    private static final int UPGRADE_DAMAGE = 3;
+
+    public Frostbite() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        baseDamage = DAMAGE;
```
dbjorge: Anything that consumes should probably have a `PERSISTENT_POSITIVE_EFFECT` tag

src/main/java/stsjorbsmod/actions/ConsumeCardIfFatalAction.java
Cartopol: A generic do-something-on-kill action would be great, but I'm not sure I know how to do that :o  Could you give me an idea of what that would look like? What do I pass into it?

dbjorge: In Java, the usual interface for a generic callback like this is called `Runnable` (if you wanted to use something that took a parameter, it's `Consumer`). Using it would look something like this:

```java
public DamageWithOnKillEffectAction(AbstractCreature target, DamageInfo info, Runnable onKillCallback) {
    // ... assign to fields
}

public void update() {
    // as HealOnFatalDamageAction, except replace the HealAction bit with
    this.onKillCallback.run();
}

// elsewhere, in each individual card...

public void use(/* ... */) {
    Runnable onKillEffect = () -> addToBot(new HealAction(p, p, heal));
    addToBot(new DamageWithOnKillEffectAction(m, new DamageInfo(p, damage), onKillEffect));
}
```

src/main/java/stsjorbsmod/actions/DamageWithOnKillEffectAction.java
```diff
@@ -0,0 +1,43 @@
...
+import com.megacrit.cardcrawl.powers.MinionPower;
+
+public class DamageWithOnKillEffectAction extends AbstractGameAction {
+    private DamageInfo info;
+    private Runnable onKillCallback;
+    private boolean worksOnMinions;
+
+
+    public DamageWithOnKillEffectAction(AbstractCreature target, DamageInfo info, Runnable onKillCallback) {
+        this.info = info;
```
dbjorge: You can use a delegating constructor here instead of repeating the block of settings (ie, this first ctor's body can be just `this(target, info, onKillCallback, false);`

src/main/java/stsjorbsmod/actions/DamageWithOnKillEffectAction.java
```diff
@@ -0,0 +1,43 @@
...
+        this.setValues(target, info);
+        this.actionType = ActionType.DAMAGE;
+        this.duration = 0.1F;
+        this.onKillCallback = onKillCallback;
+        this.worksOnMinions = worksOnMinions;
+    }
+
+    public void update() {
+        if (this.duration == 0.1F && this.target != null) {
+            this.target.damage(this.info);
```
dbjorge: This still needs to have a clearPostCombatActions check; I would recommend modelling this update method very closely after what you did for `HealOnFatalDamageAction`, I think it's exactly right if you swap out its HealAction part for the new runnable thing

src/main/java/stsjorbsmod/actions/DamageWithOnKillEffectAction.java
```diff
@@ -0,0 +1,43 @@
...
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.powers.MinionPower;
+
+public class DamageWithOnKillEffectAction extends AbstractGameAction {
+    private DamageInfo info;
+    private Runnable onKillCallback;
+    private boolean worksOnMinions;
+
+
+    public DamageWithOnKillEffectAction(AbstractCreature target, DamageInfo info, Runnable onKillCallback) {
```
dbjorge: This probably obsoletes the need for `HealOnFatalDamageAction`; maybe update Siphon to use this new class instead and delete the obsolete class

PR #272, opened 2019-12-27
Implemented Cull starter card Siphon

Cartopol:
1: Deal 10 damage. If this kills an enemy gain 5hp.

Made 2 assumptions:
- No upgrade specified, went with 4 extra damage, no extra healing.
- Only heals on non-minion enemy to reduce stalling


src/main/java/stsjorbsmod/actions/HealOnFatalAttackAction.java
```diff
@@ -0,0 +1,39 @@
...
+        this.actionType = ActionType.DAMAGE;
+        this.duration = 0.1F;
+
+    }
+
+    public void update() {
+        if (this.duration == 0.1F && this.target != null) {
+            AbstractDungeon.effectList.add(new FlashAtkImgEffect(this.target.hb.cX, this.target.hb.cY, AttackEffect.NONE));
+            this.target.damage(this.info);
+            if ((this.target.isDying || this.target.currentHealth <= 0) && !this.target.halfDead && !this.target.hasPower(MinionPower.POWER_ID)) {
```
dbjorge: My understanding is that we generally want CULL's on-kill effects to trigger even for minion kills, since it already has a character-level anti-stalling mechanic with Manifest

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -419,6 +419,10 @@
     "NAME": "Shake",
     "DESCRIPTION": "If enemy is Vulnerable, draw !M! cards. NL Apply !stsjorbsmod:UrMagic! Vulnerable. NL Exhaust."
   },
+  "stsjorbsmod:Siphon": {
+    "NAME": "Siphon",
+    "DESCRIPTION": "Deal !D! damage. NL If Fatal, heal !M! HP."
```
dbjorge: Have to avoid Fatal keyword since we want it to trigger on non-minions, unfortunately

src/main/java/stsjorbsmod/actions/HealOnFatalAttackAction.java
```diff
@@ -0,0 +1,39 @@
...
+        this.info = info;
+        this.setValues(target, info);
+        this.heal = healAmount;
+        this.actionType = ActionType.DAMAGE;
+        this.duration = 0.1F;
+
+    }
+
+    public void update() {
+        if (this.duration == 0.1F && this.target != null) {
```
dbjorge: I think for the timing of the effects it might make more sense to be consistent with `VampireDamageAction` (ie, duration of `Settings.ACTION_DUR_MED`, do the `FlashAtkImgEffect` at the start of the duration, then `tickDuration()`, then if `isDone` execute the damage + heal effects

src/main/java/stsjorbsmod/actions/HealOnFatalAttackAction.java
```diff
@@ -0,0 +1,39 @@
...
+import com.megacrit.cardcrawl.actions.AbstractGameAction;
+import com.megacrit.cardcrawl.actions.common.HealAction;
+import com.megacrit.cardcrawl.cards.DamageInfo;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.powers.MinionPower;
+import com.megacrit.cardcrawl.vfx.combat.FlashAtkImgEffect;
+
+public class HealOnFatalAttackAction extends AbstractGameAction {
```
dbjorge: I think `Damage` instead of `Attack` would be more in line with similar action names (`VampireDamageAction`, `DamageAction`, etc)

PR #273, opened 2019-12-27
Cull starter card Spirit Shield

Cartopol:
1: Reduce all damage taken by 1 until next turn.

Made 2 assumptions:
- multiple Spirit Shields stack by adding turns, not damage reduction, for its interaction with Intangible
- Upgrade was not specified. Opted for lowering cost from 1 to 0.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -176,6 +176,15 @@
       "You cannot remember any memories or clarities."
     ]
   },
+  "stsjorbsmod:SpiritShieldPower": {
+    "NAME": "Spirit Shield",
+    "DESCRIPTIONS": [
+      "Reduce all damage by ",
```
dbjorge: Prefer using String.format placeholders rather than concatenation for power localizations (I went through yesterday and updated all the existing cases)

src/main/java/stsjorbsmod/powers/SpiritShieldPower.java
```diff
@@ -0,0 +1,74 @@
...
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.localization.PowerStrings;
+import com.megacrit.cardcrawl.powers.AbstractPower;
+import stsjorbsmod.JorbsMod;
+import stsjorbsmod.util.TextureLoader;
+
+import static stsjorbsmod.JorbsMod.makePowerPath;
+
+public class SpiritShieldPower extends AbstractPower implements CloneablePowerInterface {
```
dbjorge: Prefer using `CustomJorbsModPower` and `StaticPowerInfo` to reduce boilerplate (I updated all the existing powers to use them yesterday, any of them should be good examples)

dbjorge: (this will require renaming the power images to `SpiritShield_32.png` and `SpiritShield_84.png`)

PR #274, opened 2019-12-27
Implemented Cull Uncommon card Double

Cartopol:
1: The next card (2 cards) you play this turn with Exhaust is played twice.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -40,6 +40,14 @@
       " damage to ALL enemies and lose all Coil."
     ]
   },
+  "stsjorbsmod:DoubleExhaustPower": {
+    "NAME": "Double",
+    "DESCRIPTIONS": [
+      "Your next card that #yExhausts is played twice this turn.",
+      "Your next ",
```
dbjorge: In newer powers we've started preferring using `String.Format` and placeholder strings instead of splitting the string into parts like this and concatenating, since it's more localization-friendly. ie, instead of the last 2 DESCRIPTIONS entries here, `Your next %1$s cards that #yExhaust are played twice this turn.`

PR #276, opened 2019-12-28
Implements Cull Rare card Parting Blow

Cartopol:
3: Deal 100 (120) - max hp damage.


src/main/java/stsjorbsmod/cards/cull/PartingBlow.java
```diff
@@ -0,0 +1,56 @@
...
+    public PartingBlow() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        baseDamage = 0;
+        magicNumber = baseMagicNumber = MAXIMUM_DAMAGE;
+    }
+
+    @Override
+    public int calculateBonusBaseDamage() {
+        int maxHP = AbstractDungeon.player.maxHealth;
+        return magicNumber - maxHP;
```
dbjorge: Should probably clamp this to a minimum of 0

Cartopol: From my tests it appears that it's clamped implicitly somewhere. I'm committing a clamped version for safety.

PR #278, opened 2019-12-28
Implements Cull uncommon card Expose

Cartopol:
2: Deal 30 damage. If there are still enemies alive lose 10 hp.

Currently you lose HP when killing Gremlin Leader with minions alive, even though it ends the fight.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -134,6 +134,10 @@
     "DESCRIPTION": "Apply stsjorbsmod:Slow to ALL enemies. NL Add !M! stsjorbsmod:Web into your hand.",
     "UPGRADE_DESCRIPTION": "Apply stsjorbsmod:Slow to ALL enemies. NL Add !M! stsjorbsmod:Webs into your hand."
   },
+  "stsjorbsmod:Expose": {
+    "NAME": "Expose",
+    "DESCRIPTION": "Deal !D! damage. NL If there are enemies still alive, lose !M! HP."
```
dbjorge: Since jorbs clarified that we *do* want to suppress the HP loss if there are only minions still alive, maybe we can reword this to be more clear? `If there are still enemies active`? `If any enemies are still willing to fight`? `If any enemies are still able to fight`?

src/main/java/stsjorbsmod/actions/ExposeAction.java
```diff
@@ -0,0 +1,42 @@
...
+            AbstractDungeon.effectList.add(new FlashAtkImgEffect(this.target.hb.cX, this.target.hb.cY, AttackEffect.BLUNT_HEAVY));
+        }
+        tickDuration();
+        if (isDone) {
+            this.target.damage(this.info);
+            if (AbstractDungeon.getCurrRoom().monsters.areMonstersBasicallyDead()) {
+                AbstractDungeon.actionManager.clearPostCombatActions();
+            } else {
+                AbstractPlayer p = AbstractDungeon.player;
+                addToBot(new LoseHPAction(p, p, hpLoss));
```
dbjorge: Will need an update now that Jorbs clarified that the HP loss shouldn't happen if only minions remain

src/main/java/stsjorbsmod/actions/ExposeAction.java
```diff
@@ -31,11 +33,18 @@ public void update() {
...
+            for (AbstractMonster m : AbstractDungeon.getCurrRoom().monsters.monsters) {
+                if (!m.isDeadOrEscaped() && !m.hasPower(MinionPower.POWER_ID)) {
+                    nonMinionsLeft = true;
+                }
+            }
+            if (nonMinionsLeft) {
                 AbstractPlayer p = AbstractDungeon.player;
                 addToBot(new LoseHPAction(p, p, hpLoss));
+            } else {
+                AbstractDungeon.actionManager.clearPostCombatActions();
```
dbjorge: clearPostCombatActions should probably still be under a separate `areMonstersBasicallyDead` check per usual, rather than in an else clause here

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -136,7 +136,7 @@
   },
   "stsjorbsmod:Expose": {
     "NAME": "Expose",
-    "DESCRIPTION": "Deal !D! damage. NL If there are enemies still alive, lose !M! HP."
+    "DESCRIPTION": "Deal !D! damage. NL If this does not end the fight, lose !M! HP."
```
dbjorge: I like this wording, except maybe use `combat` rather than `fight` for consistency with other card wordings

PR #285, opened 2019-12-29
Implements Cull common card Miracle:

Cartopol:
0: Gain 1(2) energy. Add 1 manifest. Retain.
Includes keyword for Manifest.


src/main/java/stsjorbsmod/cards/cull/Miracle_Cull.java
```diff
@@ -0,0 +1,46 @@
...
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        AlwaysRetainField.alwaysRetain.set(this, true);
+        magicNumber = baseMagicNumber = ENERGY_GAIN;
+        metaMagicNumber = baseMetaMagicNumber = MANIFEST_INCREASE;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster abstractMonster) {
+        addToBot(new GainEnergyAction(magicNumber));
+        ((Cull) AbstractDungeon.player).manifest += metaMagicNumber;
```
dbjorge: Even though we don't intend to officially support Prismatic Shard/etc, there are some folks that have chosen to play with other mods that similarly add new ways to get cards on mismatched classes; it's okay for us to not support the class specific mechanic in that case, but it's nice for it to at least avoid crashing, so it would be good to guard this line in an instanceof check

PR #286, opened 2019-12-29
Implements Cull common Panic

Cartopol:
0: Gain 1 Intangible. Lose 3 max hp. Ephemeral


src/main/java/stsjorbsmod/cards/cull/Panic.java
```diff
@@ -0,0 +1,50 @@
...
+        magicNumber = baseMagicNumber = INTANGIBLE_AMOUNT;
+        urMagicNumber = baseUrMagicNumber = MAXHP_LOSS;
+        EphemeralField.ephemeral.set(this, true);
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster abstractMonster) {
+        addToBot(new ApplyPowerAction(p, p, new IntangiblePlayerPower(p, magicNumber)));
+        AbstractDungeon.effectList.add(new FlashAtkImgEffect(p.hb.cX,p.hb.cY, AbstractGameAction.AttackEffect.BLUNT_LIGHT));
+        AbstractDungeon.player.decreaseMaxHealth(urMagicNumber);
```
dbjorge: I think we need to create a new LoseMaxHPAction for this to use rather than calling `decreaseMaxHealth` directly here because the power Phantasm deals damage on gaining intangible; if that damage ends the fight, we're going to want the max HP loss to be prevented (via the usual path where clearPostCombatActions would do so in the damage action).

dbjorge: (Possibly it would be good for the attack effect to be moved into that action as well)

dbjorge: I actually needed a `DecreaseMaxHpAction` for a separate purpose earlier tonight, so that's now pushed to master; you can merge master and then re-use that action rather than implementing a new one!

PR #287, opened 2019-12-29
Implements Cull rare card Pathos

Cartopol:
3: Cards in your hand cost 0 this turn. Add a curse to your discard pile for each card played.

Does not yet work for cards created in hand.


src/main/java/stsjorbsmod/powers/PathosPower.java
```diff
@@ -0,0 +1,73 @@
...
+    @Override
+    public void onUseCard(AbstractCard card, UseCardAction action) {
+        this.flash();
+        AbstractCard c = AbstractDungeon.returnRandomCurse();
+        addToBot(new MakeTempCardInDiscardAction(c, 1));
+    }
+
+    @Override
+    public void onInitialApplication() {
+        Iterator var1 = AbstractDungeon.player.hand.group.iterator();
```
dbjorge: Can shorten this a little to

```java
for (AbstractCard c : AbstractDungeon.player.hand.group) {
    c.setCostForTurn(-9);
}
```

src/main/java/stsjorbsmod/powers/PathosPower.java
```diff
@@ -0,0 +1,73 @@
...
+
+        while(var1.hasNext()) {
+            AbstractCard c = (AbstractCard)var1.next();
+            c.setCostForTurn(-9);
+        }
+    }
+
+    @Override
+    public void onCardDraw(AbstractCard card) {
```
dbjorge: I investigated how to deal with the temp cards being added to your hand in the middle of a turn; I think the most straightforward option would be to replace this onCardDraw hook usage with a new `SpirePostfixPatch` on `CardGroup::addToHand` that does the same `card.setCostForTurn(-9)`; it looked like that was going to cover all of the different temp-card-adding actions I could see.

I considered whether using `freeToPlayOnce` would be better than using `setCostForTurn(-9)`; I think what you're already doing is correct/better, since we want to avoid making X cost cards free to play (for consistency with corruption).

Cartopol: That was quite doable, thanks for the suggestion. I'm real inexperienced with patching mechanisms, so let me know if I did anything poorly.

PR #288, opened 2019-12-29
Implements Cull common card Ritual

Cartopol:
1: Gain 1 Intangible. Retain (2) target card(s) in hand until next turn. Exhaust


src/main/java/stsjorbsmod/cards/cull/Ritual.java
```diff
@@ -0,0 +1,53 @@
...
+        magicNumber = baseMagicNumber = CARDS_TO_RETAIN;
+        urMagicNumber = baseUrMagicNumber = INTANGIBLE;
+        exhaust = true;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        addToBot(new ApplyPowerAction(p, p, new IntangiblePlayerPower(p, urMagicNumber)));
+        // This makes sure that choice window does not pop-up when this is the last card played from hand
+        if (!(p.hand.size() == 1 && p.hand.group.get(0) == this) && !p.hand.isEmpty() && !AbstractDungeon.player.hasRelic(RunicPyramid.ID) && !AbstractDungeon.player.hasPower(EquilibriumPower.POWER_ID)) {
```
dbjorge: I think maybe it would be safest to move this whole if block into a separate action; that way, if there's anything that triggers a card draw somehow in response to the intangible being applied (or in response to phantasm damaging/killing something after intangible is applied), we'll properly respect the updated hand size.

PR #290, opened 2019-12-30
Made a better Frostbite upgrade

Cartopol:
- changes rarity on upgrade to Rare, so that consuming it gives a better reward. Turns it from a garbage upgrade to actually worth considering.

### DIFF
--- a/src/main/java/stsjorbsmod/cards/cull/Frostbite.java
+++ b/src/main/java/stsjorbsmod/cards/cull/Frostbite.java
@@ -39,6 +39,8 @@ public class Frostbite extends CustomJorbsModCard {
     public void upgrade() {
         if (!upgraded) {
             upgradeName();
+            this.rarity = CardRarity.RARE;
+            this.bannerImageRarity = CardRarity.RARE;
             upgradeDamage(UPGRADE_DAMAGE);
             upgradeDescription();
         }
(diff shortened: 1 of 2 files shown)


src/main/java/stsjorbsmod/cards/cull/Frostbite.java
```diff
@@ -39,6 +39,8 @@ public void use(AbstractPlayer p, AbstractMonster m) {
     public void upgrade() {
         if (!upgraded) {
             upgradeName();
+            this.rarity = CardRarity.RARE;
```
dbjorge: This is an interesting idea, but I'm not sure it'll be clear enough to a player what the impact of this will be. Maybe we can come up with a way to make the keyword more clear (maybe separate it into lines by rarity like we do for memory tooltips by remember/passive/forget effect)?

I'm a little concerned about whether players would expect this to mean that Frostbite+ should show up in the rare card pool as a possible card reward, but I don't think it's a big enough concern to block the idea.

Personally, I wouldn't spend a lot of time worrying about the upgrade effects of the starter cards; for the Wanderer, I ended up just having to replace most of what I had made up in future passes on the design spreadsheet

Cartopol: I've made a rough version of a keyword description with all the effects listed per rarity, but it's such a wall of text. I'll push it to this fork in case you want to try it. It might be better than not having it at all.

PR #291, opened 2019-12-30
Implements Cull uncommon card Seance

Cartopol:
1: Return an exhausted card to your hand, it gains ethereal. Exhaust.

- Contains an action that is a cleaned up copy of ExhumeAction which also sets isEthereal to true.
- In order to add "Ethereal. " to the card description, you pass a part of the EXTENDED_DESCRIPTION of a card into it, so that it can be localized. The result is nice, but the process feels a little dirty. Perhaps there is a cleaner place to draw the string from?
- Exhausted Seances and Exhumes cannot be brought back with Seance. Exhume is not altered, so Seance can be brought back with Exhume.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -447,6 +447,13 @@
     "NAME": "Scorching Ray",
     "DESCRIPTION": "Remember stsjorbsmod:Lust. NL Apply !stsjorbsmod:MetaMagic! stsjorbsmod:Burning !M! times. This stsjorbsmod:Burning is affected by Strength."
   },
+  "stsjorbsmod:Seance": {
+    "NAME": "Seance",
+    "DESCRIPTION": "Put a card from your exhaust pile into your hand. It gains Ethereal. NL Exhaust",
```
dbjorge: `Ethereal` should be `stsjorbsmod:Ethereal` (here and in extended desc)

dbjorge: `Exhaust` missing trailing period

dbjorge: We have an `stsjorbsmod:Exhume` keyword that could shorten this description a bit

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -447,6 +447,13 @@
     "NAME": "Scorching Ray",
     "DESCRIPTION": "Remember stsjorbsmod:Lust. NL Apply !stsjorbsmod:MetaMagic! stsjorbsmod:Burning !M! times. This stsjorbsmod:Burning is affected by Strength."
   },
+  "stsjorbsmod:Seance": {
+    "NAME": "Seance",
```
dbjorge: I think the more common spelling uses an `é`, no?

dbjorge: (Avoiding that in the id is good, but for the NAME property I think the accent is fine)

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -447,6 +447,13 @@
...
+    "DESCRIPTION": "Put a card from your exhaust pile into your hand. It gains Ethereal. NL Exhaust",
+    "EXTENDED_DESCRIPTION": [
+      " Ethereal. NL "
```
dbjorge: If this is a prefix, is the leading space necessary?

src/main/java/stsjorbsmod/actions/ExhumeAndMakeEtherealAction.java
```diff
@@ -0,0 +1,137 @@
...
+        this.setValues(this.p, AbstractDungeon.player, this.amount);
+        this.actionType = ActionType.CARD_MANIPULATION;
+        this.duration = Settings.ACTION_DUR_FAST;
+    }
+
+    public void update() {
+        Iterator<AbstractCard> c;
+        AbstractCard card;
+        if (this.duration == Settings.ACTION_DUR_FAST) {
+            if (AbstractDungeon.player.hand.size() == 10) {
```
dbjorge: This should use `BaseMod.MAX_HAND_SIZE` rather than 10 (BaseMod patches the original ExhumeAction to account for this)

src/main/java/stsjorbsmod/actions/ExhumeAndMakeEtherealAction.java
```diff
@@ -0,0 +1,137 @@
...
+                    this.isDone = true;
+                } else {
+                    AbstractDungeon.gridSelectScreen.open(this.p.exhaustPile, 1, TEXT[0], false);
+                    this.tickDuration();
+                }
+            }
+
+        } else {
+            if (!AbstractDungeon.gridSelectScreen.selectedCards.isEmpty()) {
+                for (c = AbstractDungeon.gridSelectScreen.selectedCards.iterator(); c.hasNext(); card.unhover()) {
```
dbjorge: For relatively complex classes like `ExhumeAction`, the decompiler built into IntelliJ isn't very good at producing particularly readable output. I find that [JD-GUI](https://java-decompiler.github.io/) produces output that is much more readable (eg, it doesn't have this confusing `card.unhover()`-as-the-for-loop-incrementor thing in its output, and it also does a better job of decompiling to for-each loops instead of weird iterator nesting patterns); I would recommend starting from that when you need to re-use large parts of a decompiled class

src/main/java/stsjorbsmod/actions/ExhumeAndMakeEtherealAction.java
```diff
@@ -0,0 +1,137 @@
...
+public class ExhumeAndMakeEtherealAction extends AbstractGameAction {
+    private AbstractPlayer p;
+    private static final UIStrings uiStrings;
+    public static final String[] TEXT;
+    private String prependDescription;
+
+    private ArrayList<AbstractCard> exhumes = new ArrayList<>();
+
+
+    public ExhumeAndMakeEtherealAction(String prependDescription) {
```
dbjorge: I think it would make more sense for this class to contain the responsibility of setting the description prefix (using a UI string, see `IfEnemyIntendsToAttackAction` for example), rather than accepting it as a constructor param

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
Cartopol: I believe you're thinking of Ephemeral. According to the spec sheet, this should actually become Ethereal, but maybe Ephemeral makes more sense?

src/main/java/stsjorbsmod/actions/ExhumeAndMakeEtherealAction.java
Cartopol: I was wondering why the code looked so shoddy! I had no idea the decompiler had such a large impact. I'll be using JD-GUI from now on. 
Do you mind if I leave this one as is for now, considering your idea of patching this into the existing Exhume action?

Cartopol: I've rebuilt it anyway from the JD-GUI decompiled ExhumeAction. I've made only the necessary adjustments to it (added Séance to excluded cards, set Ethereal property on card, add string to card description). Oh, and I removed the unused boolean upgrade property from the action.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
dbjorge: Oh, I'm sorry, you're right; I was thinking of Ephemeral

PR #292, opened 2019-12-30
Implements Cull uncommon card Abjure

Cartopol:
0: Put all Spirit Shields in your deck into your hand. (Retain)

- According to sheet, this should draw Spirit Shield from 'everywhere'
- implemented as: Draw all Spirit Shields from draw pile, discard pile and exhaust pile


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -1,4 +1,9 @@
 {
+  "stsjorbsmod:Abjure": {
+    "NAME": "Abjure",
+    "DESCRIPTION": "Draw all Spirit Shields from your draw pile, discard pile and exhaust pile.",
```
dbjorge: I'm not sure `Draw` is the right word to use here, since I don't think this would actually trigger "on draw" effects from other powers/relics

src/main/java/stsjorbsmod/cards/cull/Abjure.java
```diff
@@ -0,0 +1,55 @@
...
+            if (c instanceof SpiritShield_Cull) {
+                addToBot(new ExhaustToHandAction(c));
+            }
+        }
+    }
+
+    @Override
+    public void upgrade() {
+        if (!upgraded) {
+            this.selfRetain = true;
```
dbjorge: `selfRetain` is currently only implemented in the beta branch, so until that releases, we're using the (equivalent) Until the beta branch releases, we're using `AlwaysRetainField.alwaysRetain.set(this, true);`

src/main/java/stsjorbsmod/actions/GetCardFromDrawPileAction.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public GetCardFromDrawPileAction(AbstractCard card) {
+        this.actionType = ActionType.CARD_MANIPULATION;
+        this.card = card;
+        this.duration = Settings.ACTION_DUR_FAST;
+    }
+
+    public void update() {
+        if (this.duration == Settings.ACTION_DUR_FAST) {
+            if (AbstractDungeon.player.drawPile.contains(this.card) && AbstractDungeon.player.hand.size() < 10) {
```
dbjorge: `BaseMod.MAX_HAND_SIZE` rather than 10

src/main/java/stsjorbsmod/cards/cull/Abjure.java
```diff
@@ -0,0 +1,55 @@
...
+        for (AbstractCard c : p.discardPile.group) {
+            if (c instanceof SpiritShield_Cull) {
+                addToBot(new DiscardToHandAction(c));
+            }
+        }
+        for (AbstractCard c : p.exhaustPile.group) {
+            if (c instanceof SpiritShield_Cull) {
+                addToBot(new ExhaustToHandAction(c));
+            }
+        }
```
dbjorge: Consider adding a `p.createHandIsFullDialog()` at the end if there were more eligible cards than there was space in hand

src/main/java/stsjorbsmod/actions/GetCardFromDrawPileAction.java
```diff
@@ -0,0 +1,37 @@
...
+                this.card.applyPowers();
+                AbstractDungeon.player.drawPile.removeCard(this.card);
+            }
+
+            AbstractDungeon.player.hand.refreshHandLayout();
+            AbstractDungeon.player.hand.glowCheck();
+        }
+
+        this.tickDuration();
+        this.isDone = true;
```
dbjorge: `isDone = true;` makes `tickDuration()` and the initial `if(this.duration ...)` check both unnecessary.

Issue #243, opened 2019-12-15
burning: UI element that shows shield on fire if an enemy has block


wang429:
Provide more clarity to Burning when owner has burning and block. Maybe set the shield on fire or a partial bar on the shield.

Cartopol:
Since merging the burning block icon, the block icon is always burning, even if block's owner is not burning. Presumably the RenderBurningBlockPatch needs to include a check on having BurningPower before rendering the block icon.

PR #294, opened 2019-12-31
Adds a fire icon behind the block icon

wang429:
If there's block to show and the creature is burning, show fire icon behind the block icon
image
addresses #243

### DIFF
--- a/src/main/java/stsjorbsmod/patches/RenderBurningBlockPatch.java
+++ b/src/main/java/stsjorbsmod/patches/RenderBurningBlockPatch.java
@@ -3,38 +3,28 @@ package stsjorbsmod.patches;
 import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.g2d.SpriteBatch;
 import com.evacipated.cardcrawl.modthespire.lib.SpirePatch;
+import com.evacipated.cardcrawl.modthespire.lib.SpirePrefixPatch;
 import com.megacrit.cardcrawl.core.AbstractCreature;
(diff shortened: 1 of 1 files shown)


src/main/java/stsjorbsmod/patches/RenderBurningBlockPatch.java
```diff
@@ -0,0 +1,40 @@
...
+import javassist.CannotCompileException;
+import javassist.expr.ExprEditor;
+import javassist.expr.MethodCall;
+import stsjorbsmod.JorbsMod;
+import stsjorbsmod.powers.BurningPower;
+
+public class RenderBurningBlockPatch {
+    @SpirePatch(clz = AbstractCreature.class, method = "renderHealth")
+    public static class AbstractCreature_renderHealthPatch {
+        public static ExprEditor Instrument() {
```
dbjorge: It feels like this would be much simpler as a prefix patch on `renderBlockIconAndValue`, no?

```java
@SpirePrefixPatch
public static void patch(AbstractCreature __this, SpriteBatch sb, float x, float y) {
    if (__this.hasPower(BurningPower.POWER_ID)) {
        sb.setColor(Color.GOLD.cpy());
        sb.draw(
            ImageMaster.ATK_FIRE,
            x + AbstractCreature.BLOCK_ICON_X - 32.0F,
            y + AbstractCreature.BLOCK_ICON_Y - 32.0F + __this.blockOffset,
            32.0F,
            32.0F,
            64.0F,
            64.0F,
            Settings.scale,
            Settings.scale,
            0.0F);
    }
}
```

PR #303, opened 2020-01-01
Implements Cull rare Sentinel

Cartopol:
1: Deal 10(13) damage. Lower Manifest counter by 1. Exhaust.

- adds functionality to IncreaseManifestAction so that 'increase' can be negative.
- IncreaseManifestAction now takes into account its amount property instead of always incrementing by 1
- order of operations in Sentinel's use() reversed, otherwise Manifest decrease would be removed by ClearPostCombatActions if this card ends combat


src/main/java/stsjorbsmod/actions/IncreaseManifestAction.java
```diff
@@ -34,14 +32,18 @@ public void update() {
         Cull player = (Cull) target;
 
         if (this.duration == this.startDuration) {
-            String msg = String.format(TEXT[0], amount);
+
+            String msg = amount < 0 ?  String.format(TEXT[1], amount) : String.format(TEXT[0], amount);
             AbstractDungeon.effectsQueue.add(new TextAboveCreatureEffect(target.hb.cX - target.animX, target.hb.cY, msg, Settings.RED_TEXT_COLOR));
```
dbjorge: Should probably also make the color of the effect green if the amount is negative

src/main/java/stsjorbsmod/cards/cull/Sentinel_Cull.java
```diff
@@ -0,0 +1,48 @@
...
+    public Sentinel_Cull() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        baseDamage = DAMAGE;
+        magicNumber = baseMagicNumber = MANIFEST_DECREASE;
+        this.exhaust = true;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        addToBot(new IncreaseManifestAction(-magicNumber));
```
dbjorge: Would be good to leave a comment noting why the ordering is important

PR #311, opened 2020-01-04
move fire to between block icon and block number

wang429:
I suppose this could be an Insert patch, but this just felt less painful for some reason. Probably because it was easier to just grab an initial implementation, copy paste, and move some stuff around than it was to re-remember writing a Locator

Also, should the block icon should be a slightly different color? 
image


wang429:
Yeah I like it better redder.

wang429:
Using the burning power icon behind the block.
Icon spins behind block because it amused me.
Block icon color shifts closer to red the higher the block percent is affected by burning. Never more than 50% between the light blue and red though.

src/main/java/stsjorbsmod/patches/RenderBurningBlockPatch.java
```diff
@@ -27,8 +36,47 @@ public static void patch(AbstractCreature __this, SpriteBatch sb, float x, float
                         64.0F,
                         Settings.scale,
                         Settings.scale,
-                        0.0F);
+                        burningRotation);
+                burningRotation += 0.1;
```
dbjorge: I wouldn't recommend incrementing by a constant value every re-render like this; there's a lot of variation in how long it takes between render calls, both between different users/machines and different frames on the same machine. I also wouldn't recommend tying it to a single static variable like this, since that will make the rotation speed dependent on the number of burning creatures this is rendering for.

I would recommend the pattern that `AbstractMemory` uses for rotating the active memory effect. The pattern looks like:
* store a rotation timer as a field on the entity in question (here, as a `SpireField` on `AbstractCreature`
* on each *update* tick (not each render), do something like this:
  ```java
    private void updateYourRotationTimer(AbstractCreature c) {
        yourRotationField.set(c, yourRotationField.get(c) - Gdx.graphics.getDeltaTime());
        if (yourRotationField.get(c) < 0.0F) {
            yourRotationField.set(c, YOUR_ROTATION_DURATION);
        }
    }
  ```
* In the render path, use a value of `(yourRotationField.get(c) / YOUR_ROTATION_DURATION) * 360F` for the actual rotation amount to pass to the drawing routines

src/main/java/stsjorbsmod/patches/RenderBurningBlockPatch.java
```diff
@@ -27,8 +36,47 @@ public static void patch(AbstractCreature __this, SpriteBatch sb, float x, float
...
             }
         }
+
+        public static ExprEditor Instrument() {
+            return new ExprEditor() {
+//                @Override
```
dbjorge: Best to just remove the commented code, that's what version control is for :)

src/main/java/stsjorbsmod/patches/RenderBurningBlockPatch.java
```diff
@@ -27,8 +36,47 @@ public static void patch(AbstractCreature __this, SpriteBatch sb, float x, float
...
+                    if (f.getClassName().contains(AbstractCreature.class.getName()) && f.getFieldName().equals("blockColor")) {
+                        f.replace(String.format("{ $_ = %1$s.renderBurningBlock(this, $proceed()); }",
+                                RenderBurningBlockPatch.class.getName()));
+                    }
+                }
+            };
+        }
+    }
+
+    public static void renderFireOnBlock(SpriteBatch sb, float x, float y, float BLOCK_ICON_X, float BLOCK_ICON_Y, float blockOffset) {
```
dbjorge: I think this can go away along with the commented version of the code

src/main/java/stsjorbsmod/patches/RenderBurningBlockPatch.java
wang429: Admittedly, the rotation thing was wholly for my amusement. I'm not totally convinced animating the power icon makes much sense. I'll see what I can do though.

PR #312, opened 2020-01-04
Implements Cull uncommon card Catharsis

Cartopol:
1: (Power) You can play curses (costs 1 life). Deal 5 damage to a random enemy whenever you play a curse for each curse played so far this combat.

- note: 'Curses played this combat' is counted in the amount2 of CatharsisPower. If any curses were played previously (ie. with Blue Candle), these will not be counted.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -33,6 +33,13 @@
       "Mmmm... Escargot"
     ]
   },
+  "stsjorbsmod:CatharsisPower": {
+    "NAME": "Catharsis",
+    "DESCRIPTIONS": [
+      "You can play Curses. Whenever you do, lose #b%3$s HP and deal #b%1$s damage to ALL enemies for each Curse played this combat ( #b%2$s Curses played).",
+      "You can play Curses. Whenever you do, lose #b%3$s HP and deal #b%1$s damage to ALL enemies for each Curse played this combat ( #b%2$s Curse played)."
```
wang429: IMO, the power description should more closely match Blue Candle's "Unplayable Curse cards can now be played. Whenever you play a Curse, lose # hp and Exhaust it".

dbjorge: I agree that it should note the "unplayable" part; I don't think it should affect Pride, in particular

src/main/java/stsjorbsmod/patches/PlayableCursesPatch.java
```diff
@@ -0,0 +1,26 @@
...
+    @SpirePatch(
+            clz = AbstractCard.class,
+            method = "canUse"
+    )
+    public static class AbstractCard_canUse
+    {
+        @SpirePrefixPatch
+        public static SpireReturn<Boolean> patch(AbstractCard __this)
+        {
+            if (__this.type == AbstractCard.CardType.CURSE && AbstractDungeon.player.hasPower(CatharsisPower.POWER_ID)) {
```
dbjorge: I think this is going to interact incorrectly with Pride (showing you as being able to play it when you don't have enough energy). Rather than unilaterally returning true here, it might be better to use an instrument patch that replaces calls to `AbstractDungeon.player.hasRelic("Blue Candle")` with `(AbstractDungeon.player.hasRelic("Blue Candle") || AbstractDungeon.player.hasPower(CatharsisPower.POWER_ID))`.

The instrument patch machinery can't see at the time of the patch that the argument to hasRelic is the constant string "Blue Candle", so you have to be a little trickier in the patch replacement; the pattern looks like:

```java
@Override
public void edit(MethodCall methodCall) throws CannotCompileException {
    if (methodCall.getClassName().equals(AbstractPlayer.class.getName()) &&
        methodCall.getMethodName().equals("hasRelic"))
    {
                "{ $_ = ($proceed($$) || (%1$s.isBlueCandle($1) && %1$s.hasCatharsisPower($0))); }",
                PlayableCursesPatch.class.getName()));
    }
}
```

src/main/java/stsjorbsmod/powers/CatharsisPower.java
```diff
@@ -0,0 +1,59 @@
...
+    public static final String POWER_ID = STATIC.ID;
+
+    private static final int HP_LOSS = 1;
+
+    public CatharsisPower(final AbstractCreature owner, final int damagePerCurse) {
+        super(STATIC);
+
+        this.owner = owner;
+        this.amount = damagePerCurse;
+        this.amount2 = 0;
```
dbjorge: I'm not sure tracking as amount2 is ideal here; I think the wording of the power suggests that it should be counting cases where you played curses with blue candle/pride before playing the power. I think maybe querying `AbstractDungeon.actionManager.cardsPlayedThisCombat` would be a better way to go

PR #313, opened 2020-01-04
Implements two Cull changes regarding Legendary cards

Cartopol:
- Splinter Soul ignores all Legendary cards in draw pile when finding a Consume target
- Spirit Shield now has Legendary tag

### DIFF
--- a/src/main/java/stsjorbsmod/cards/cull/SplinterSoul.java
+++ b/src/main/java/stsjorbsmod/cards/cull/SplinterSoul.java
@@ -37,10 +37,9 @@ public class SplinterSoul extends CustomJorbsModCard {
     public void use(AbstractPlayer p, AbstractMonster m) {
         addToBot(new DamageAction(m, new DamageInfo(p, damage), AttackEffect.SLASH_VERTICAL));
 
-        CardGroup drawPileWithoutLegendary = p.drawPile.getPurgeableCards();
-
-        if (!drawPileWithoutLegendary.isEmpty()) {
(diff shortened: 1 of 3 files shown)


src/main/java/stsjorbsmod/cards/cull/SplinterSoul.java
```diff
@@ -36,8 +37,10 @@ public SplinterSoul() {
     public void use(AbstractPlayer p, AbstractMonster m) {
         addToBot(new DamageAction(m, new DamageInfo(p, damage), AttackEffect.SLASH_VERTICAL));
 
-        if (!p.drawPile.isEmpty()) {
-            AbstractCard randomCardInDrawPile = p.drawPile.getRandomCard(AbstractDungeon.cardRandomRng);
+        CardGroup drawPileWithoutLegendary = p.drawPile.getPurgeableCards();
```
wang429: super minor, but technically, this would also exclude bottled cards.

dbjorge: +1; we should also consider whether it's a good idea or not for it to exclude Ascender's Bane and Necronomicurse. If we do support bottled cards, we should double check the interaction with the bottle relic, it may be necessary to add clearing associated bottle relics out to the destroy card helper

Cartopol: According to Jorbs on discord, bottled cards and unremovable curses should not be eligible for consumption. getPurgeableCards gets rid of the unremovable curses. Wrapping it in a CardGroup.getGroupWithoutBottledCards also removes the bottled ones. 
Currently this is implemented in SplinterSoul's use method, since it is the only current one which consumes other cards. We could consider generalizing this elsewhere.

PR #316, opened 2020-01-05
Implements Cull uncommon card Channeling

Cartopol:
1: Play the top card of your deck for each incoming attack. Exhaust.

- Uses ReflectionHacks to get access to private properties of AbstractMonster. Not sure if this is desired.


src/main/java/stsjorbsmod/cards/cull/Channeling.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster abstractMonster) {
+        for (int i = 0; i < magicNumber; ++i) {
+            this.addToBot(new PlayTopCardAction(AbstractDungeon.getCurrRoom().monsters.getRandomMonster(null, true, AbstractDungeon.cardRandomRng), false));
+        }
+    }
+
+    @Override
+    public String getRawDynamicDescriptionSuffix() {
+        return magicNumber == 1 ? EXTENDED_DESCRIPTION[0] : EXTENDED_DESCRIPTION[1];
```
wang429: For localization purposes, this assumes that other languages don't have special pluralization forms for 0 and 2, which is false. I'm not sure we've done the best job with localization language compatibility anyways. Might not be totally important right now.

dbjorge: Sadly, neither we nor the base game do a usable job of handling languages with non-english-like pluralization forms

dbjorge: The card description probably shouldn't reveal this much information if the player has runic dome; maybe have this check for `IntentUtils.areIntentsHidden()` and return empty string in that case

src/main/java/stsjorbsmod/cards/cull/Channeling.java
```diff
@@ -0,0 +1,72 @@
...
+    private static final int COST = 1;
+
+    public Channeling() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        this.exhaust = true;
+        baseMagicNumber = 0;
+    }
+
+    private boolean isMonsterAttacking(AbstractMonster m) {
+        return m.intent.equals(AbstractMonster.Intent.ATTACK) ||
```
dbjorge: Can replace this with `IntentUtils.isAttackIntent(m.intent)`

src/main/java/stsjorbsmod/cards/cull/Channeling.java
```diff
@@ -0,0 +1,72 @@
...
+import basemod.ReflectionHacks;
+import com.megacrit.cardcrawl.actions.common.PlayTopCardAction;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import stsjorbsmod.JorbsMod;
+import stsjorbsmod.cards.CustomJorbsModCard;
+import stsjorbsmod.characters.Cull;
+
+public class Channeling extends CustomJorbsModCard {
```
dbjorge: Would be nice to add

```java
@Override
public boolean shouldGlowGold() {
    return IntentUtils.playerCanSeeThatAnyEnemyIntentMatches(IntentUtils::isAttackIntent);
}
```

src/main/java/stsjorbsmod/cards/cull/Channeling.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public int calculateBonusMagicNumber() {
+        int channelCount = 0;
+        for (AbstractMonster m : AbstractDungeon.getCurrRoom().monsters.monsters) {
+            if (!m.isDying && m.currentHealth > 0 && !m.isEscaping) {
+                if (isMonsterAttacking(m)) {
+                    if ((boolean)ReflectionHacks.getPrivate(m, AbstractMonster.class, "isMultiDmg")) {
+                        channelCount += (int)ReflectionHacks.getPrivate(m, AbstractMonster.class, "intentMultiAmt");
```
dbjorge: This is a nice touch, I hadn't considered how this should interact with multiattacks when I initially read the card description but I think this is a good interaction.

PR #317, opened 2020-01-05
Implements Cull Legendary card Mirror

Cartopol:
3: Until your next turn you take no damage and enemies deal damage to themselves. Ephemeral.

- implemented similar to ThornsPower, but instead of having DamageType.THORNS, has DamageType.NORMAL. This causes Byrds to lose flight, Writhing Mass to trigger Malleable etc. Also increases in damage automatically when the player is Vulnerable.


wang429:
It would be nice for Mirror to cause a recalculation of the damage intent amount

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -362,6 +362,10 @@
     "NAME": "Miracle",
     "DESCRIPTION": "Retain. NL Gain !M! [E] . NL Add !stsjorbsmod:MetaMagic! stsjorbsmod:Manifest."
   },
+  "stsjorbsmod:Mirror": {
+    "NAME": "Mirror",
+    "DESCRIPTION": "stsjorbsmod:Ephemeral. NL Until your next turn, take no damage. NL All enemies attack themselves."
```
dbjorge: I think the `All enemies attack themselves` wording is maybe a little off; my reading of it would imply that it should retarget non-damage parts of attacks with power side effects, too, but I don't think that's going to happen. Maybe `Until your next turn, enemies that would damage you instead damage themselves.` or something?

dbjorge: Actually, the original wording from the sheet seems fine here (`Until your next turn, you take no damage and enemies deal damage to themselves.`)

src/main/java/stsjorbsmod/powers/MirrorPower.java
```diff
@@ -0,0 +1,58 @@
...
+        } else {
+            AbstractDungeon.actionManager.addToBottom(new ReducePowerAction(this.owner, this.owner, this.ID, 1));
+        }
+    }
+
+    @Override
+    public int onAttacked(DamageInfo info, int damageAmount) {
+        if (info.type != DamageInfo.DamageType.THORNS && info.type != DamageInfo.DamageType.HP_LOSS && info.owner != null && info.owner != this.owner) {
+            this.flash();
+            this.addToTop(new DamageAction(info.owner, new DamageInfo(this.owner, damageAmount, DamageInfo.DamageType.NORMAL), AbstractGameAction.AttackEffect.SLASH_HORIZONTAL, false));
```
dbjorge: This is going to cause the damage amount to be affected by the amount of vulnerable/intangible/etc on the player, and not by similar effects on the monster; I'm not sure whether this is desirable or not (it feels consistent with the idea of a "mirror" but maybe inconsistent with the current card description). Should maybe ask for design feedback about how these interactions should work

src/main/java/stsjorbsmod/powers/MirrorPower.java
```diff
@@ -0,0 +1,58 @@
...
+    public void atStartOfTurn() {
+        if (this.amount == 1) {
+            AbstractDungeon.actionManager.addToBottom(new RemoveSpecificPowerAction(this.owner, this.owner, this.ID));
+        } else {
+            AbstractDungeon.actionManager.addToBottom(new ReducePowerAction(this.owner, this.owner, this.ID, 1));
+        }
+    }
+
+    @Override
+    public int onAttacked(DamageInfo info, int damageAmount) {
```
dbjorge: When we needed a similar effect for Mirror Image, I got feedback that it was weird for the attack animation to still show on the player when it was being redirected; you can avoid that by using `OnDamageToRedirectSubscriber` like Mirror Image does, eg:

```java
    @Override
    public boolean onDamageToRedirect(AbstractPlayer player, DamageInfo info, AttackEffect attackEffect) {
        if (info.type != DamageInfo.DamageType.THORNS && info.type != DamageInfo.DamageType.HP_LOSS && info.owner != null && info.owner != this.owner) {
            this.flash();
            this.addToTop(new DamageAction(info.owner, new DamageInfo(this.owner, info.output, DamageInfo.DamageType.NORMAL), attackEffect, false));
            return true;
        }
        return false;
    }
```

src/main/java/stsjorbsmod/powers/MirrorPower.java
Cartopol: This was answered by jorbs on Discord, saying that it should account for the Vulnerable and Intangible on the enemies. However, I'm a little stuck how to implement this. 
I'm trying to build a patch in the AbstractMonster.damage method, to block it from setting the info.output to 1 when you are Intangible. So I wanted to replace the hasPower("IntangiblePlayer") condition like so:
```
public void edit(MethodCall methodCall) throws CannotCompileException {
    if (methodCall.getMethodName().equals("hasPower")) {
                "{ $_ = ($proceed($$) && !(%1$s.isIntangiblePlayerPower($1) && %1$s.playerHasMirrorPower() )); }",
                IgnorePlayerIntangiblePatch.class.getName()));
    }
}
```

But it's never hitting the loggers that I added in isIntangiblePlayerPower() and playerHasMirrorPower(). Could you help me out with how to patch this in? I'd love to get it working but I'm not sure how :)

dbjorge: I think the particular issue you're hitting here is that AbstractPlayer overrides AbstractMonster::damage, so you'd want to be patching AbstractPlayer rather than AbstractMonster for it to take effect.

That said, it looks like there's maybe been some re-thinking in discord... from earlier tonight,

```
[12:00 AM] brosmike: "Mirror: 3: Until your next turn you take no damage and enemies deal damage to themselves. Ephemeral." If you have both mirror and intangible, say, should the attack intent number above enemies say "1" or "the amount of damage the enemy is about to do to itself"? Do we need to come up with some different icon to replace the normal red sword intent image to make what's happening less confusing?
[12:05 AM] brosmike: In particular, if the player has intangible and mirror, and an enemy is vulnerable and intends to perform a 10 damage attack, is it least-confusing for the damage intent to read as "1", "10", or "15"? I lean towards "15"
[12:18 AM] jorbs: oh mmm
[12:18 AM] jorbs: i think it's meant to not work with intangible
[12:18 AM] jorbs: so they'd only deal themselves 1
[12:18 AM] jorbs: at least for right now
[12:18 AM] brosmike: does that imply the damage modifiers for both the player and the enemy should apply?
[12:19 AM] brosmike: ie, if both player and enemy have vulnerable, does vulnerable apply twice?
[12:19 AM] brosmike: or is intangible special?
[12:20 AM] brosmike: I hope you're still consuming as much scotch as I am while we consider this question
[12:22 AM] jorbs: it deals thorns-type damage back to the enemy
[12:25 AM] brosmike: So, enemy damage/damage intent numbers are unchanged from usual and account for player's powers, and the only enemy stuff that matters is stuff that would apply to thorns-type-damage (block, intangible, etc), and any such enemy thing applies in addition to and after the player powers.
[12:25 AM] brosmike: That's straightforward enough to implement
[12:28 AM] brosmike: Is the "you take no damage" part of the description actually intended to prevent all damage, or just normal-type damage from enemies? ie, should it affect thorns damage the player would take, and/or hp-loss-type-damage (eg, from playing a curse with blue candle)
[12:44 AM] jorbs: all
```

dbjorge: Note that the OnDamageToRedirectSubscriber patch already operates a little earlier than onAttacked, so definitely recommend looking at that switch before investigating further patches

PR #320, opened 2020-01-06
Implements Cull potion Ghost Poison

Cartopol:
Gain 1 manifest for each enemy in combat, then win the combat if non-elite/non-boss.

- ends combat using m.die() on all monsters notDeadOrEscaping.

### DIFF
--- a/src/main/java/stsjorbsmod/potions/GhostPoisonPotion.java
+++ b/src/main/java/stsjorbsmod/potions/GhostPoisonPotion.java
@@ -1,0 +1,78 @@
+package stsjorbsmod.potions;
+
+import com.megacrit.cardcrawl.actions.animations.VFXAction;
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.helpers.PowerTip;
+import com.megacrit.cardcrawl.localization.PotionStrings;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import com.megacrit.cardcrawl.potions.AbstractPotion;
+import com.megacrit.cardcrawl.powers.BackAttackPower;
+import com.megacrit.cardcrawl.rooms.AbstractRoom;
+import com.megacrit.cardcrawl.vfx.combat.SmokeBombEffect;
+import stsjorbsmod.JorbsMod;
+import stsjorbsmod.actions.IncreaseManifestAction;
+
+public class GhostPoisonPotion extends AbstractPotion {
+    public static final String POTION_ID = JorbsMod.makeID(GhostPoisonPotion.class.getSimpleName());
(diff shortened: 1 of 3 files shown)


src/main/java/stsjorbsmod/potions/GhostPoisonPotion.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public void use(AbstractCreature target) {
+        int liveMonsterCount = 0;
+
+        for (AbstractMonster m : (AbstractDungeon.getCurrRoom()).monsters.monsters) {
+            if (!m.isDeadOrEscaped()) {
+                ++liveMonsterCount;
+                addToBot(new VFXAction(new SmokeBombEffect(m.hb.cX, m.hb.cY)));
+                m.die();
```
dbjorge: This strategy isn't compatible with the `3_Darklings` fight; using the potion with a half-dead darkling causes it to still add manifest/hide health bar/disable end turn button, but it doesn't actually end the fight, it just softlocks

src/main/java/stsjorbsmod/potions/GhostPoisonPotion.java
```diff
@@ -0,0 +1,78 @@
...
+    private static final PotionStrings potionStrings = CardCrawlGame.languagePack.getPotionString(POTION_ID);
+    public static final String NAME = potionStrings.NAME;
+    public static final String[] DESCRIPTIONS = potionStrings.DESCRIPTIONS;
+
+    private static final int BASE_POTENCY = 1;
+
+    public GhostPoisonPotion() {
+        super(NAME, POTION_ID, PotionRarity.COMMON, PotionSize.H, PotionColor.SMOKE);
+        this.potency = this.getPotency();
+        this.description = String.format(DESCRIPTIONS[0], this.potency);
```
dbjorge: Note that potency is affected by Sacred Bark, but the current `use` implementation is not; I think it's a funny interaction for sacred bark to double the manifest this potion causes you to use, but regardless of whether that interaction is intended or not, the description and effect should be consistent

src/main/java/stsjorbsmod/potions/GhostPoisonPotion.java
```diff
@@ -0,0 +1,78 @@
...
+        this.isThrown = true;
+        this.tips.add(new PowerTip(this.name, this.description));
```
dbjorge: Consider also adding a tip for the manifest keyword

Cartopol: I'm not sure how to reference a keyword for these purposes, since they're not a class. Considering other potions don't clarify using tips, I hope that's ok for now.

src/main/java/stsjorbsmod/potions/GhostPoisonPotion.java
Cartopol: I've removed the usage of potency for this potion. It's probably too much of an unwelcome surprise to have Sacred Bark screw you over like this. If anything, I think we should consider a positive interaction with Sacred Bark, such as gaining a static 1 Manifest, instead of 1 per enemy.

PR #328, opened 2020-01-27
Bottled Memory Relic

wang429:
Slight issue with using the CustomBottleRelic implementation; tiny bottle icon when rendering everywhere.
image
image


src/main/java/stsjorbsmod/patches/BottledMemoryPatch.java
```diff
@@ -0,0 +1,113 @@
...
+    @SpirePatch(clz = CardGroup.class, method = "initializeDeck", paramtypez = {CardGroup.class})
+    public static class AbstractGroup_initializeDeck {
+        public static ExprEditor Instrument() {
+            return new ExprEditor() {
+                @Override
+                public void edit(FieldAccess fieldAccess) throws CannotCompileException {
+                    // Transforms `!inBottleTornado` into `!(inBottleTornado || inBottleMemory)`
+                    // Logically equivalent to `!inBottleTornado && !inBottleMemory`
+                    if (fieldAccess.getClassName().equals(AbstractCard.class.getName()) && fieldAccess.getFieldName().equals("inBottleTornado")) {
+                        fieldAccess.replace("{ $_ = ($proceed() || stsjorbsmod.patches.BottledMemoryPatch.isInBottleMemory(c)); }");
```
wang429: I preferred following pre-existing bottled relic draw patterns rather the example in the DefaultModBase

dbjorge:
Filed https://github.com/daviscook477/BaseMod/issues/243 for the tiny icon on the card

src/main/java/stsjorbsmod/patches/BottledMemoryPatch.java
```diff
@@ -0,0 +1,113 @@
...
+
+        @Override
+        public void edit(MethodCall methodCall) throws CannotCompileException {
+            if (methodCall.getClassName().equals(CardGroup.class.getName()) && methodCall.getMethodName().equals(methodName)) {
+                methodCall.replace("{ $_ = (stsjorbsmod.patches.BottledMemoryPatch.removeBottledCards($proceed())); }");
+            }
+        }
+    }
+
+    /**
```
dbjorge: This comment doesn't add any information that the method signature doesn't already cover

src/main/java/stsjorbsmod/patches/BottledMemoryPatch.java
```diff
@@ -0,0 +1,113 @@
...
+import com.megacrit.cardcrawl.relics.BottledFlame;
+import com.megacrit.cardcrawl.relics.BottledLightning;
+import com.megacrit.cardcrawl.relics.BottledTornado;
+import javassist.CannotCompileException;
+import javassist.expr.ExprEditor;
+import javassist.expr.FieldAccess;
+import javassist.expr.MethodCall;
+import stsjorbsmod.relics.BottledMemoryRelic;
+
+public class BottledMemoryPatch {
```
dbjorge: Based on searching for other base game references to `inBottleFlame`/etc, I think we're going to want a few more patches:

* [ ] `Duplicator`, `DollysMirror`: Need to special case these to reset the `inBottleMemory` field back to false after they duplicate their target card
* [ ] `FountainOfCurseRemoval`: Need to special case to avoid picking bottle cards

wang429: Sort of pretending that the FountainOfCurseRemoval actually did the right thing. ran through 250ish floors without seeing it, gave up.

PR #330, opened 2020-01-28
Coda

wang429:
1: Deal 24(30) damage. If this would kill an enemy instead it puts the enemy at 1hp.

I took some liberties with description wording. May need to adjust it.


src/main/java/stsjorbsmod/patches/CodaPatch.java
```diff
@@ -0,0 +1,36 @@
...
+        }
+    }
+
+    @SpirePatch(clz = AbstractMonster.class, method = "damage")
+    public static class AbstractMonster_damage {
+        public static ExprEditor Instrument() {
+            return new ExprEditor(){
+                @Override
+                public void edit(MethodCall methodCall) throws CannotCompileException {
+                    if (methodCall.getClassName().equals(AbstractMonster.class.getName()) && methodCall.getMethodName().equals("healthBarUpdatedEvent")) {
```
dbjorge: I think it would probably be a little friendlier to other effects to leave `this.lastDamageTaken` (set higher up than `healthBarUpdatedEvent` in `AbstractMonster::damage`) accurate to what Coda ends up dealing; I'd recommend patching in right before the field access to `lastDamageTaken` instead to check for hpRetained and clamp to `this.currentHealth - hpRetained.get(info)`

PR #331, opened 2020-01-30
Damnation

wang429:


src/main/java/stsjorbsmod/powers/DamnationPower.java
```diff
@@ -0,0 +1,45 @@
...
+
+        updateDescription();
+    }
+
+    @Override
+    public void onUseCard(AbstractCard card, UseCardAction action) {
+        CardMetaUtils.destroyCardPermanently(card);
+        AbstractDungeon.transformCard(card, isAutoUpgrade, AbstractDungeon.miscRng);
+        AbstractDungeon.topLevelEffectsQueue.add(new ShowCardAndObtainEffect(AbstractDungeon.getTransformedCard(), Settings.WIDTH / 2.0F, Settings.HEIGHT / 2.0F, false));
+        addToBot(new RemoveSpecificPowerAction(owner, owner, this.ID));
```
dbjorge: This should probably add to top; if there's a queue of card uses queued somehow, we don't want the power applying to several of them

src/main/java/stsjorbsmod/powers/DamnationPower.java
```diff
@@ -0,0 +1,45 @@
...
+        super(STATIC);
+
+        this.owner = owner;
+        this.isAutoUpgrade = isAutoUpgrade;
+
+        updateDescription();
+    }
+
+    @Override
+    public void onUseCard(AbstractCard card, UseCardAction action) {
```
dbjorge: Should we exclude cards that aren't in the master deck? If not, should we at least exclude transient copies of cards used by "play next card twice" type mechanics? (Probably need to ask discord)

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -46,6 +46,14 @@
       "Whenever you remember a new memory, deal #b%1$s damage to ALL enemies and lose all Coil."
     ]
   },
+  "stsjorbsmod:DamnationPower": {
+    "NAME": "Damnation",
+    "DESCRIPTIONS": [
+      "Permanently #yTransform the next card played this combat",
```
dbjorge: It's more localization friendly to just have 2 versions of the whole sentence written out, like in the card strings; it's okay for it to be a little redundant, it makes it easier for languages where the more natural phrasing might put the then clause before the object of the sentence

src/main/java/stsjorbsmod/powers/DamnationPower.java
wang429: Still asking about cards that aren't in the master deck

wang429: cards not in the master deck will consume the charge without transforming

src/main/java/stsjorbsmod/powers/DamnationPower.java
```diff
@@ -0,0 +1,48 @@
...
+            CardMetaUtils.destroyCardPermanently(card);
+            AbstractDungeon.transformCard(card, isAutoUpgrade, AbstractDungeon.miscRng);
+            AbstractDungeon.topLevelEffectsQueue.add(new ShowCardAndObtainEffect(AbstractDungeon.getTransformedCard(), Settings.WIDTH / 2.0F, Settings.HEIGHT / 2.0F, false));
+        }
+        addToTop(new RemoveSpecificPowerAction(owner, owner, this.ID));
+    }
+
+    @Override
+    public void updateDescription() {
+        description = isAutoUpgrade ? DESCRIPTIONS[0] : DESCRIPTIONS[1];
```
dbjorge: I think this is inverted from how the power strings are ordered, no?

PR #346, opened 2020-02-06
balance 20200205

wang429:
Purge mind upgrade 11 -> 9.
Trauma damage 32 -> 28.
Trauma+ damage 37 -> 33
Lust no longer applies burning on remember.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Memory-Strings.json
```diff
@@ -50,7 +50,7 @@
   "stsjorbsmod:LustMemory": {
     "NAME": "Lust",
     "DESCRIPTIONS": [
-      "#bRemember: Apply !R! #yBurning to ALL enemies. NL #bPassive: Your attacks apply !P! #yBurning."
+      "#bPassive: Your attacks apply !P! #yBurning."
```
dbjorge: Keyword string for also needs an update

Issue #352, opened 2020-02-06
Remembering Patience should not trigger coil


wang429:
Put another way: 

> coil blows up whenever you remember a non-patience memory (instead of a new memory)

Issue #350, opened 2020-02-06
take "gain coil" off snake oil


wang429:

dbjorge:
From tonight:

> [12:14 AM] jorbs: let's try: snake oil: remember patience, deal damage to all enemies equal to your coil
> [12:15 AM] jorbs: or whatever the right wording is

PR #364, opened 2020-02-08
Snake oil + Coil

wang429:
Snake oil remembers patience
Snake oil does not add 4 coil
Snake oil damage does not include playing itself

Coil does not trigger when remember patience

addresses #350 and #352

### DIFF
--- a/src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Keyword-Strings.json
+++ b/src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Keyword-Strings.json
@@ -17,7 +17,7 @@
   },
   {
     "NAMES": ["coil"],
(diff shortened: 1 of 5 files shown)


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Keyword-Strings.json
```diff
@@ -17,7 +17,7 @@
   },
   {
     "NAMES": ["coil"],
-    "DESCRIPTION": "Gain 1 Coil each time you play a card while remembering #yPatience. When remembering new memories or ending your turn while snapped, Coil is consumed to deal damage to ALL enemies."
+    "DESCRIPTION": "Gain 1 Coil each time you play a card while remembering #yPatience. When you remember new, #ynon-Patience memories or ending your turn while snapped, Coil is consumed to deal damage to ALL enemies."
```
dbjorge: The clause `When you remember new, #ynon-Patience memories or ending your turn while snapped` is awkward; "you remember" and "ending" shouldn't be different verb tenses.

Issue #348, opened 2020-02-06
Grimoire relic


wang429:
Give the grimoire relic a turn counter.

PR #366, opened 2020-02-08
add turn counter to grimoire relic

wang429:
addresses #348


src/main/java/stsjorbsmod/relics/GrimoireRelic.java
```diff
@@ -25,6 +25,12 @@ public void atBattleStart() {
         AbstractPlayer p = AbstractDungeon.player;
         this.flash();
         AbstractDungeon.actionManager.addToBottom(new RememberSpecificMemoryAction(p, PatienceMemory.STATIC.ID));
+        this.counter = 0;
+    }
+
+    @Override
+    public void atTurnStart() {
+        this.counter++;
```
dbjorge: I'm not sure we want this to be incrementing with Quicksilver; I asked for feedback about the interaction in discord, we might want to change this reflect `AbstractDungeon.actionManager.turn` instead of maintaining its own count

wang429: I have a change ready for this, but I'll wait for a response before pushing it.

wang429: Thinking about this a bit more, I realize that Grimoire the relic and Grimoire the card have nothing to do with each other. I suppose using the same Grimoire power asset that just says, "at the start of turn 7, exhume grimoire" would be enough?

wang429: Thoughts on making the new power to handle the self-exhume on snap and on turn 7?

src/main/java/stsjorbsmod/characters/Wanderer.java
```diff
@@ -449,6 +450,12 @@ public void onVictory() {
     public void preBattlePrep() {
         super.preBattlePrep();
         snapCounter.reset();
+        HashSet<AbstractCard> allInstances = GetAllInBattleInstances.get(masterDeck.group.stream().filter(c -> c.cardID.equals(ForbiddenGrimoire.ID)).findAny().get().uuid);
```
dbjorge: I don't think making it a Wanderer character inherent is the way to go, here; we've already gotten feedback about folks hoping for cards/etc to work if forced into other characters, and while I don't think we should bend over backwards to make that work perfect, I also would prefer to avoid regressing existing behavior that worked perfectly fine for those cases

src/main/java/stsjorbsmod/cards/wanderer/ForbiddenGrimoire.java
```diff
@@ -27,11 +27,12 @@
     private static final int COST = 0;
     private static final int CARD_PLAYS_TO_EXHUME = 3;
 
+    public static final int EXHUME_TURN = 7;
+
     public ForbiddenGrimoire() {
         super(ID, COST, TYPE, COLOR, RARITY, TARGET);
         magicNumber = baseMagicNumber = CARD_PLAYS_TO_EXHUME;
         EntombedField.entombed.set(this, true);
-        SelfExhumeFields.selfExhumeAtStartOfTurn7.set(this, true);
```
dbjorge: Was removing this intentional? It doesn't look like the new marker power has this effect

src/main/java/stsjorbsmod/characters/Wanderer.java
dbjorge: (my preference would be to add an `onEntomb()` hook to either `CustomJorbsModCard` or a new subscriber interface, call it from `EntombedPatch.AbstractPlayer_preBattlePrep`, and implement that on the `ForbiddenGrimoire` card to implement the marker power)

src/main/java/stsjorbsmod/cards/wanderer/ForbiddenGrimoire.java
wang429: Probably one or two too many ctrl-z's 🤦‍♂

wang429:
Exhuming Grimoire on turn 7 is now handled by the power.

EntombedGrimoirePower
EntombedGrimoirePowerSnap

Issue #361, opened 2020-02-07
Sacred bark does not double liquid virtue


wang429:
**Version**
Mod version: PUT 0.5? HERE (and if it isn't the latest version, try to reproduce there)

> Neablis Today at 12:45 AM
So I took sacred bark purely to test the interactions with mod potions. (it should double the effects of all potions)
It does not double liquid virtue

wang429:
Steps to reproduce:
1. add Liquid Virtue potion
2. add Sacred Bark relic
3. hover Liquid Virtue potion, notice text still says "two"
4. while in combat, use Liquid Virtue potion.
5. notice that only two virtues clarified

Adding Sacred Bark then the potion does not have this issue.
This also effects other potions

PR #376, opened 2020-02-08
move potency, description, and tips to initializeData

wang429:
addresses #361


src/main/java/stsjorbsmod/potions/LiquidVirtue.java
```diff
@@ -23,10 +23,14 @@
...
-        this.description = DESCRIPTIONS[0];
-        this.potency = getPotency();
         this.isThrown = false;
         this.targetRequired = false;
+    }
+
+    @Override
+    public void initializeData() {
+        this.potency = getPotency();
+        this.description = DESCRIPTIONS[0];
```
dbjorge: Description should be dynamic to reflect potency, like BurningPotion

Issue #344, opened 2020-02-06
Make Brambles more like Claw


wang429:
Should be similar to Defect's Claw

dbjorge:
Update Brambles to be "0: Deal damage equal to the number of Material Components played this combat", upgrade to "3 + number of components"

PR #383, opened 2020-02-10
Brambles damage increases with material components played

wang429:
addresses #344 

image

### DIFF
--- a/src/main/java/stsjorbsmod/cards/wanderer/materialcomponents/MaterialComponentsDeck.java
+++ b/src/main/java/stsjorbsmod/cards/wanderer/materialcomponents/MaterialComponentsDeck.java
@@ -5,7 +5,6 @@ import com.megacrit.cardcrawl.cards.CardGroup;
 import com.megacrit.cardcrawl.cards.CardGroup.CardGroupType;
 
 import java.util.ArrayList;
-import java.util.HashSet;
 
 /*
 Material Components Deck:	Contains 7 common, 6 uncommon, 3 rare Material Components cards.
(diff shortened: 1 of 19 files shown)


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -58,7 +58,8 @@
   },
   "stsjorbsmod:Brambles": {
     "NAME": "Brambles",
-    "DESCRIPTION": "Deal !D! damage."
+    "DESCRIPTION": "Deal damage equal to the number of stsjorbsmod:Material\u00a0Components played this combat. NL (Deals !D! damage)",
```
dbjorge: For the `(Deals !D! damage.)` suffix, take a look at how `Mania` uses `getRawDynamicDescriptionSuffix` and `EXTENDED_DESCRIPTION`; doing it that way avoids having the parenthetical present in the master deck and card selection screens, where it's more confusing than helpful.

src/main/java/stsjorbsmod/cards/wanderer/materialcomponents/MaterialComponentsDeck.java
```diff
@@ -16,6 +16,8 @@
 	resets completely for next fight.
  */
 public class MaterialComponentsDeck {
+    public static int playedCount = 0;
```
dbjorge: It would be nice if this name were a little less ambiguous about "since when" - maybe `playedThisCombatCount` or something?

Issue #226, opened 2019-12-09
Wrath count on cards overlaps with Bottle.


wang429:
Wrath count icon is in the same spot as the bottle would be. We should move this elsewhere, I'm thinking below the energy icon.

![image](https://user-images.githubusercontent.com/1097069/70401470-b60dcf00-19fd-11ea-91ac-408dacc5efb1.png)
related to comment by Frenoss in #139

wang429:
Gonna move to the left under energy; probably the most intuitive.

wang429:
Not specific to wanderer, the CULL has wrath too.

PR #391, opened 2020-02-11
Wrath icon relocated to under energy

wang429:
Addresses #226 

image


dbjorge:
Tried a few variations of this and posted results in discord. Feelings were mixed on whether folks prefer the old vs new positioning. Leaving the PR open while we gather more feedback in discord.

dbjorge:
Note, I just merged in a new `wrath` dev console command to make this easier to test, so if you play with this some more you'll want to merge master to get that

wang429:
Given that jorbs seems to really prefer the top left, under the energy icon route, I think I'll leave this as is. That said, I did just notice that I could have just moved the icon in `card_overlay_wrath_icon_512` rather than doing the math... Probably the better approach to the problem.
Also, would be nice to have a higher resolution wrath icon. If it also happens to be smaller, then Wrath icon could probably fit a bit higher and more snug with energy.

PR #393, opened 2020-02-11
Thirst campfire option

Jacob-Laverty:
Initial implementation of Thirst mechanic. 

Completely replaces the Rest option for Cull with the Thirst mechanic as defined in the spreadsheet. 

The confirmation screen could probably stand to have some better UX patched into it to show exact heal amounts for a selected but not confirmed card however that can come later.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-UI-Strings.json
```diff
@@ -71,6 +71,18 @@
       "At the end of turn 7, #ySnap. NL (It is turn #b%1$s now.)"
     ]
   },
+  "stsjorbsmod:ThirstOption": {
+    "TEXT": [
+      "Thirst",
+      "Consume wrath from an attack, heal 5 hp, and gain 1 max hp per wrath",
```
dbjorge: `#yWrath`? (not sure if smart text works here)

dbjorge: I feel like this phrasing is a little ambiguous about whether the heal 5 is per-stack or not. Maybe something like `Consume ALL Wrath from a card. For each Wrath consumed, heal 5 HP and gain 1 Max HP.`?

src/main/java/stsjorbsmod/patches/CampfireThirstPatch.java
```diff
@@ -0,0 +1,59 @@
...
+    public CampfireThirstPatch() {
+    }
+
+    public static ExprEditor Instrument() {
+       return new ExprEditor() {
+           public void edit(NewExpr e) throws CannotCompileException {
+               // Completely replace the rest option with thirst for Cull
+               if(e.getClassName().equals("com.megacrit.cardcrawl.ui.campfire.RestOption")) {
+                    e.replace(String.format(
+                            "{$_ = %1$s.getCullCampfireOption();}",
```
dbjorge: It might be friendlier to other mods that want to patch the same place to do something like `"{$_ = %1$s.getCullCampfireOption($proceed($$));}"` and return the argument from `getCullCampfireOption` in the base case; that way, another mod that happens to occur after ours with an instrument/insert patch looking for the same `new RestOption` expression will still find it

src/main/java/stsjorbsmod/campfire/ThirstOption.java
```diff
@@ -0,0 +1,35 @@
...
+public class ThirstOption extends AbstractCampfireOption {
+    public static final String[] TEXT;
+
+    public ThirstOption(boolean active) {
+        this.label = TEXT[0];
+        this.usable = active;
+        this.updateUsability(active);
+    }
+
+    public void updateUsability(boolean canUse) {
```
dbjorge: Should we be making this option unusable if you somehow reach it without having any cards with wrath stacks?

src/main/java/stsjorbsmod/campfire/ThirstOption.java
```diff
@@ -0,0 +1,35 @@
...
+        this.description = canUse ? TEXT[1] : TEXT[2];
+        // TODO: Replace with big boy art
+        this.img = ImageMaster.loadImage("images/ui/campfire/ritual.png");
```
dbjorge: This is probably a little too often to be doing file IO; better would be to call `loadImage` once to initialize a static field with the Texture, then have this be `this.img = BUTTON_IMAGE;`.

src/main/java/stsjorbsmod/effects/CampfireThirstEffect.java
```diff
@@ -0,0 +1,110 @@
...
+        if (!AbstractDungeon.isScreenUp) {
+            this.duration -= Gdx.graphics.getDeltaTime();
+            this.updateBlackScreenColor();
+        }
+
+        Iterator var1;
+        if (!AbstractDungeon.isScreenUp && !AbstractDungeon.gridSelectScreen.selectedCards.isEmpty()) {
+            var1 = AbstractDungeon.gridSelectScreen.selectedCards.iterator();
+
+            while(var1.hasNext()) {
```
dbjorge: Can simplify a few lines of iterator stuff here down to `for (AbstractCard c : AbstractDungeon.gridSelectScreen.selectedCards) {`

I'm guessing this was probably copied from IntelliJ's decompilation of `CampfireSmithEffect` - generally, if you're using a decompiled base game class as a starting point, I'd recommend using [JD-GUI](https://java-decompiler.github.io/) to do the decompilation instead of trusting IntelliJ; IntelliJ is convenient to F12 with, but JD-GUI has more accurate and more conventional output (it uses a for-each loop in this spot in `CampfireSmithEffect`, for example)

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-UI-Strings.json
wang429: If smart test works, I also recommend prepending `#b` to the numbers.

dbjorge: Smart text doesn't work, unfortunately

Issue #385, opened 2020-02-10
Fear On Spire Spear Does Not Turn Wanderer Around


johnplunk:
Used Fear card on Spire spear, after it fled the character model for Wanderer did not auto-turn around.

Back Attack was still active until the model did actually turn around

![image](https://user-images.githubusercontent.com/47401869/74128557-0858a100-4ba3-11ea-9371-64bd589ad404.png)

PR #394, opened 2020-02-11
if the spear is no longer here because of fear, face its dear

wang429:
addresses #385 

This should work if a fight exists with multiple monsters on either side of the player. That is not tested though. Works in the 4 scenarios of Shield and Spear fight though.

### DIFF
--- a/src/main/java/stsjorbsmod/powers/FearPower.java
+++ b/src/main/java/stsjorbsmod/powers/FearPower.java
@@ -1,6 +1,7 @@
 package stsjorbsmod.powers;
 
 import com.megacrit.cardcrawl.actions.common.EscapeAction;
+import com.megacrit.cardcrawl.actions.common.RemoveSpecificPowerAction;
 import com.megacrit.cardcrawl.characters.AbstractPlayer;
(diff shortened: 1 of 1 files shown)


src/main/java/stsjorbsmod/powers/FearPower.java
```diff
@@ -44,7 +49,17 @@ public void atStartOfTurn() { // note: monster's turn, not player's
...
+                        ++remainingMonsters;
+                        // negative if monster to the left of player, positive if monster to the right of player
+                        monsterDisplacementsFromPlayer.add(m.drawX - AbstractDungeon.player.drawX);
+                    }
+                }
+                // flipHorizontal = false -> face right
+                // flipHorizontal = true  -> face left
+                if (monsterDisplacementsFromPlayer.stream().noneMatch(x -> x < 0 == AbstractDungeon.player.flipHorizontal)) {
+                    // If none of the remaining monsters are on the same side the player is facing, then turn
+                    AbstractDungeon.player.flipHorizontal = !AbstractDungeon.player.flipHorizontal;
```
dbjorge: This probably also wants to also lose SurroundedPower in the spire + spear casse

PR #396, opened 2020-02-11
Block/Dex -> Dmg/Str mechanic initial implementation.

Jacob-Laverty:
Kept it really simple for a first pass at this mechanic. As I was testing it, it seemed incredibly strong with things like, Kunai, Orichalcum, Ornamental Fan etc so I didn't want to spend a ton of time on details without letting people feel it out.


src/main/java/stsjorbsmod/patches/CullBlockPatch.java
```diff
@@ -0,0 +1,36 @@
...
+import com.megacrit.cardcrawl.actions.AbstractGameAction;
+import com.megacrit.cardcrawl.actions.common.DamageRandomEnemyAction;
+import com.megacrit.cardcrawl.cards.DamageInfo;
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import stsjorbsmod.characters.Cull;
+
+
+/*
+This is an easy enough first pass for Cull not doing block things. A future iteration will probably want to patch
```
dbjorge: I agree with this comment and appreciate that you've left it :)

src/main/java/stsjorbsmod/patches/CullDexPatch.java
```diff
@@ -0,0 +1,55 @@
...
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.powers.*;
+import javassist.CannotCompileException;
+import javassist.expr.ExprEditor;
+import javassist.expr.FieldAccess;
+import javassist.expr.NewExpr;
+import stsjorbsmod.characters.Cull;
+
+@SpirePatch(
```
dbjorge: I think the mechanics of the patch seem reasonable, but I would suggest putting it as a prefix patch on `update` (wrapped in a `if (duration == startingDuration)`) rather than an instrument patch on this specific CTOR; historically, there have been a few instances of the base game introducing new CTOR overloads (possibly changing which CTOR is the "true" base one you'd want this patch to apply to), but it's very unlikely that the signature of `AbstractPower.update()` will change
much more common for future changes to the base game to introduce new CTOR overloads than

PR #400, opened 2020-02-12
Balance pass 20200211

wang429:
- Grimoire returns every 4 cards after playing
- Prepare gains block before gaining coil
- Find Familiar is a start of turn power

### DIFF
--- a/src/main/java/stsjorbsmod/cards/wanderer/Prepare.java
+++ b/src/main/java/stsjorbsmod/cards/wanderer/Prepare.java
@@ -53,19 +53,6 @@ public class Prepare extends CustomJorbsModCard {
     public void use(AbstractPlayer p, AbstractMonster m) {
         addToBot(new GainBlockAction(p, p, block));
         addToBot(new ApplyPowerAction(p, p, new CoilPower(p, metaMagicNumber)));
-
-        // TODO: clean up later if we keep block -> gain coil
-        // We can't use a GainBlockAction directly because we need to recalculate block in applyPowers in between the
-        // new Coil stacks applying, this card being played (and possibly adding a new coil stack for that), and the
-        // GainBlockAction being scheduled.
-//        addToBot(new AbstractGameAction() {
-//            @Override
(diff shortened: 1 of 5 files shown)


src/main/java/stsjorbsmod/cards/wanderer/Prepare.java
```diff
@@ -51,19 +51,21 @@ public String getRawDynamicDescriptionSuffix() {
 
     @Override
     public void use(AbstractPlayer p, AbstractMonster m) {
+        addToBot(new GainBlockAction(p, p, block));
         addToBot(new ApplyPowerAction(p, p, new CoilPower(p, metaMagicNumber)));
 
+        // TODO: clean up later if we keep block -> gain coil
```
dbjorge: It's okay to just remove the obsolete code; if we need it back later, we have version control for that :)

src/main/java/stsjorbsmod/powers/FindFamiliarPower.java
```diff
@@ -16,9 +16,9 @@ public FindFamiliarPower(final AbstractCreature owner) {
     }
 
     @Override
-    public void atEndOfTurn(boolean isPlayer) {
+    public void atStartOfTurn() {
         MemoryManager mm = MemoryManager.forPlayer(owner);
-        if (isPlayer && mm != null) {
+        if (owner.isPlayer && mm != null) {
```
dbjorge: Can remove the `owner.isPlayer` here

src/main/java/stsjorbsmod/powers/FindFamiliarPower.java
```diff
@@ -16,9 +16,9 @@ public FindFamiliarPower(final AbstractCreature owner) {
     }
 
     @Override
-    public void atEndOfTurn(boolean isPlayer) {
+    public void atStartOfTurn() {
```
dbjorge: This probably needs a corresponding power/card string update

PR #421, opened 2020-02-16
explosive power now shouldn't trigger if the exploder dies over the c…

wang429:
…ourse of time eddy

maybe instead of this, addToBot in each of the `forEachApplicable*` method?

### DIFF
--- a/src/main/java/stsjorbsmod/actions/TimeEddyAction.java
+++ b/src/main/java/stsjorbsmod/actions/TimeEddyAction.java
@@ -66,7 +66,8 @@ public class TimeEddyAction extends AbstractGameAction {
         forEachApplicableMonsterPower(p -> p.atStartOfTurnPostDraw());
 
         // Monster turn
-        forEachApplicableMonsterPower(p -> p.duringTurn());
+        // This causes any actions added to the bottom as part of duringTurn to occur after all the others. Maybe not a big deal
+        addToBot(new ConsumerGameAction<>(o -> forEachApplicableMonsterPower(p -> p.duringTurn()), null));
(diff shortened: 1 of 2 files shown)


src/main/java/stsjorbsmod/actions/TimeEddyAction.java
```diff
@@ -66,7 +66,8 @@ public void update() {
         forEachApplicableMonsterPower(p -> p.atStartOfTurnPostDraw());
 
         // Monster turn
-        forEachApplicableMonsterPower(p -> p.duringTurn());
+        // This causes any actions added to the bottom as part of duringTurn to occur after all the others. Maybe not a big deal
+        addToBot(new ConsumerGameAction<>(o -> forEachApplicableMonsterPower(p -> p.duringTurn()), null));
```
dbjorge: I'm not super comfortable with this approach; even if we were to update the rest of the lines to be more consistent, that would risk new problems with the second TimeEddyAction (in that any actions added to bottom by one of the callbacks would start running after the second TImeEddyAction runs again).

wang429: Yeah, I agree. I'm not terribly happy about this approach either. It feels incredibly jank.

wang429: Another idea I had was to patch `AbstractMonster::die` to remove actions from the action manager where `this` as the owner

dbjorge:
Obsoleted by fee1235c5dcb0f06e667ad46879959a7df766b8f

PR #423, opened 2020-02-17
Ray of Frost Effect

wang429:
Similar to Scorching Ray

RayOfFrostEffect


src/main/java/stsjorbsmod/actions/RayOfFrostAction.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+
+    @Override
+    public void update() {
+        CardCrawlGame.sound.play("ATTACK_MAGIC_SLOW_1", 0.7F);
+        AbstractDungeon.effectList.add(new BorderFlashEffect(Color.SKY.cpy()));
+        AbstractDungeon.effectList.add(new ScalingLaserEffect(source.hb.cX, source.hb.cY, target.hb.cX, target.hb.cY, Color.SKY.cpy(), Color.CYAN.cpy(), damageInfo.output));
+        AbstractDungeon.effectList.add(new FlashAtkImgEffect(target.hb.cX, target.hb.cY, AttackEffect.BLUNT_LIGHT, true));
+        target.damage(damageInfo);
```
dbjorge: I think it would be friendlier to other patches (eg, those from other mods) to try to re-use the normal DamageAction where possible, rather than re-implement the damage + clearPostCombatActions check. I would lean towards handling this more like SearingBlow, where the card's `use` queues both a normal `DamageAction` and also a separate `VfxAction` for the custom `SearingBlowEffect`

PR #433, opened 2020-02-21
Display snap damage

### LINKED ISSUE #409: Render Damage to be dealt by snapping
Seems like a neat little QoL thing to do. Suggested by AYAYA(base username: Archlife#7684) on Discord
> AYAYA Yesterday at 7:53 PM
Suggestion: on turn 7 color a portion of your health bar (similar to poison) to represent the amount of health snapping will lose you. I feel like it would be nice ui, and help communicate to newer players the losing health mechanic at a glance

Will require more fiddling than just adding HealthBarRenderPower since SnapCounter isn't a power.

wang429:
displays expected snap self-damage on HP bar
some refactoring and opening up of `SnapCounter`

Fixes #409


wang429:
Using the Fragile power now also needs retinkering with Determination applying the Fragile power. It should not duplicate like Corruption, but it should also update the amount to the minimum of turn count and Determination stacks.

wang429:
This changes the behavior of playing Determination multiple times. In my opinion, it is more in line with the wording of the card, "At the end of this turn, Snap."

wang429:
Zero clarities
image

Three clarities turn 1
image

Three clarities turn 7 (snap turn)
image

All clarities turn 7 (snap turn)
image

Fragile Power + Determination
FragilePower

src/main/java/stsjorbsmod/powers/FragilePower.java
```diff
@@ -16,23 +19,41 @@ public FragilePower(final AbstractCreature owner, final int turnsUntilSnap) {
...
         } else {
-            AbstractDungeon.actionManager.addToBottom(new ReducePowerAction(this.owner, this.owner, this.ID, 1));
+            addToBot(new ReducePowerAction(this.owner, this.owner, this.ID, 1));
         }
     }
 
     @Override
     public void updateDescription() {
-        description = String.format(this.amount == 1 ? DESCRIPTIONS[0] : DESCRIPTIONS[1], this.amount);
+        description = String.format(this.amount == 1 ? DESCRIPTIONS[0] : DESCRIPTIONS[1], this.amount, amount2 * 3, amount2 * 6);
```
dbjorge: Better to make public and reuse the constants in SnapAction than repeating the 3/6 magic numbers here

src/main/java/stsjorbsmod/powers/FragilePower.java
```diff
@@ -16,23 +19,41 @@ public FragilePower(final AbstractCreature owner, final int turnsUntilSnap) {
 
         this.owner = owner;
         this.amount = turnsUntilSnap;
+        this.amount2 = MemoryManager.forPlayer(owner).countCurrentClarities();
+        this.type = SPECIAL;
```
dbjorge: I feel uneasy about using the power bar to represent this; in the past when we used a relic the feedback we got regularly was "it's confusing for quicksilver to not be affecting the 7-turn count on the 'relic'", and I would expect to get the same feedback about time eddy not interacting with Fragile.

src/main/java/stsjorbsmod/patches/HealthBarRenderSnapPatch.java
```diff
@@ -0,0 +1,64 @@
...
+import com.evacipated.cardcrawl.modthespire.lib.*;
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.helpers.ImageMaster;
+import com.megacrit.cardcrawl.powers.IntangiblePower;
+import javassist.CtBehavior;
+import stsjorbsmod.characters.Wanderer;
+import stsjorbsmod.memories.MemoryManager;
+import stsjorbsmod.powers.FragilePower;
+
+public class HealthBarRenderSnapPatch {
```
dbjorge: I'm not sure I understand why patching the patch is necessary; can we not make `FragilePower` a `HealthBarRendererPower` instead, something like this?

```java
    @Override
    public int getHealthBarAmount() {
        if (this.amount != 1) { return 0; }
        int selfDamage = amount2 * SnapAction.PLAYER_DAMAGE_PER_CLARITY;
        if (owner.hasPower(IntangiblePlayerPower.POWER_ID)) {
            selfDamage = MathUtils.clamp(selfDamage, 0, 1);
        }
        return selfDamage;
    }

    @Override
    public Color getColor() {
        return new Color(
                (MathUtils.cosDeg((float) (System.currentTimeMillis() / 10L % 360L)) + 1.25F) / 2.3F,
                (MathUtils.cosDeg((float) ((System.currentTimeMillis() + 1000L) / 10L % 360L)) + 1.25F) / 2.3F,
                (MathUtils.cosDeg((float) ((System.currentTimeMillis() + 2000L) / 10L % 360L)) + 1.25F) / 2.3F,
                1.0F
        );
    }
```

src/main/java/stsjorbsmod/powers/FragilePower.java
```diff
@@ -16,23 +19,41 @@ public FragilePower(final AbstractCreature owner, final int turnsUntilSnap) {
 
         this.owner = owner;
         this.amount = turnsUntilSnap;
+        this.amount2 = MemoryManager.forPlayer(owner).countCurrentClarities();
```
dbjorge: Seeing the "1" there as the number of clarities without any tooltip text on the power explaining what the number was or how it related to the numbers in the tooltip felt confusing to me; I think it would be better to not use an amount2 at all

src/main/java/stsjorbsmod/cards/wanderer/Determination.java
```diff
@@ -37,7 +38,17 @@ public Determination() {
     public void use(AbstractPlayer p, AbstractMonster m) {
         addToBot(new RememberSpecificMemoryAction(p, PrideMemory.STATIC.ID));
         if (!upgraded) {
-            addToBot(new ApplyPowerAction(p, p, new FragilePower(p, magicNumber), magicNumber));
+            if (p.hasPower(FragilePower.POWER_ID)) {
```
dbjorge: I suspect we'd need this check to move to an action, rather than happening within the card's `use`, to avoid a "quicksilver + determination" scenario that starts without fragile from adding fragile twice

src/main/java/stsjorbsmod/cards/wanderer/Determination.java
```diff
@@ -37,7 +38,17 @@ public Determination() {
...
+                FragilePower fragilePower = (FragilePower) p.getPower(FragilePower.POWER_ID);
+                fragilePower.flash();
+                fragilePower.amount = 1;
```
dbjorge: `magicNumber` rather than 1?

src/main/java/stsjorbsmod/cards/wanderer/Determination.java
```diff
@@ -37,7 +38,17 @@ public Determination() {
...
+                fragilePower.updateDescription();
+            } else if (p.hasPower(SnappedPower.POWER_ID)) {
+                p.getPower(SnappedPower.POWER_ID).flash();
+            } else if (!p.hasPower(FragilePower.POWER_ID) && !p.hasPower(SnappedPower.POWER_ID)) {
+                // apply FragilePower only if the player doesn't already have FragilePower or SnappedPower.
```
dbjorge: This comment isn't adding anything; what it's saying is already clear from the code

src/main/java/stsjorbsmod/powers/FragilePower.java
dbjorge: Maybe a compromise here (especially now that there's a tutorial tip pointing to the snap counter UI) could be to still use the snap counter (instead of a power icon) as the UI primary, but use it consistently for Determination as well? ie, make Determination cause the snap counter to go rainbow immediately and special case the snap counter to say "at end of your turn" in the Determination case?

I think the health bar update and the dynamic tooltip are both good regardless of the power icon, so maybe keep those somehow (either with patches against the SnapCounter or with an InvisiblePower)? It's just the power icon in the buff/debuff bar that concerns me; I expect putting an icon there to cause players to get confused about Time Eddy not applying to it.

src/main/java/stsjorbsmod/patches/HealthBarRenderSnapPatch.java
wang429: Yeah, you're right. I believe the reason I had patched the patch initially was because I wasn't using `FragilePower` initially and then just didn't rethink the approach.

src/main/java/stsjorbsmod/powers/FragilePower.java
wang429: Sure, I'll use a different variable.

src/main/java/stsjorbsmod/cards/wanderer/Determination.java
wang429: Ah, good catch.

wang429:
Brief summary of changes (from what I remember):
- new SNAP_TURN constant.
- FragilePower now doesn't show any numbers
- FragilePower now takes in a SnapCounter for the purposes of checking or forcing snap turn
- when Determination is played, the new DeterminationAction forces a SnapCounter to the snap turn, either from the existing SnapCounter or a new SnapCounter instance.
- some simple review comments

Issue #435, opened 2020-02-22
Prayer wheel gave me duplicate legendary cards


MartinRykfors:
Slay the spire: Beta
Mod version: 1.4.0

After finishing a normal fight with the Prayer Wheel (Normal enemies drop an additional card reward), both of the card choices contained Patron. I was allowed to take both copies of it.

Thanks for making this mod, it is really fun!

wang429:
Fully Reproduced using seed `205K8XCB4COCT`, Asc 2, Floor 1

wang429:
I guess I should have also mentioned that I used the following to trim the card pool size down
```
        cards = Arrays.asList(
                new ForbiddenGrimoire(),
                new Strike_Wanderer(),
                new Defend_Wanderer(),
                new EyeOfTheStorm(),
                new FreshAdventure(),
                new Patron(),
                new FaerieFire(),
                new ColorSpray(),
                new FlameWard(),
                new ThirstingSword(),
                new Entangle(),
                new MageArmor(),
                new RayOfFrost(),
                new Banish(),
                new CorrodingBarrier(),
                new ArcaneWeapon(),
                new DisguiseSelf(),
                new ArcaneForm(),
                new ChatteringStrike());
```

wang429:
Addressed by #440

PR #440, opened 2020-02-27
Remove legendary cards from the pool when they're shown

wang429:
addresses #435 
See comments for how I reproduced the issue.


dbjorge:
For a more targeted fix that sticks strictly to addressing #435 without a broader design change, probably the easiest option would be to postfix patch `setupItemRewards` to scan `this.rewards` for multiple `RewardItems` with duplicate legendary cards, and replace such `RewardItem`s with `new RewardItem()`s until there aren't duplicates anymore. If any are replaced/rerolled, will probably want to call `this.positionRewards()` again afterwards.

wang429:
My understanding of Legendary was that a player could not _see_ multiple copies of a Legendary (unless previously Destroyed in the run). That might be an outdated notion. I'll check.

wang429:
> jorbsToday at 8:09 PM
you're meant to be able to see legendaries over and over again until you pick them up, at which point they stop being findable unless you get rid of them somehow

PR #450, opened 2020-03-05
Implements new Cull card: Commune

Cartopol:
0: Scry 4 (6)


wang429:
Something that might be a fun little easter egg is referencing defect's old reprogram in some way.

PR #452, opened 2020-03-06
Implements new Cull card: Paralysis

Cartopol:
3(2): Retain. Gain 1 Intangible. Exhaust. Costs 1 less each turn it is in your hand.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -418,6 +418,10 @@
     "NAME": "Panic",
     "DESCRIPTION": "stsjorbsmod:Ephemeral. NL Gain !M! Intangible. NL Lose !stsjorbsmod:UrMagic! Max HP."
   },
+  "stsjorbsmod:Paralysis": {
+    "NAME": "Paralysis",
+    "DESCRIPTION": "Retain. NL Gain !M! Intangible. NL When Retained, lower its cost by 1 this combat. NL Exhaust."
```
dbjorge: I don't think cards normally implicitly refer to themselves with "its", and the closest equivalent (Establishment) uses "reduce cost" rather than "lower cost", should probably be consistent. `When retained, reduce this card's cost by 1 this combat.`?

Cartopol: This wording is taking from Sands of Time, which is the closest equivalent. "Retain. NL Deal !D! damage. NL When Retained, lower its cost by 1 this combat."

PR #454, opened 2020-03-07
rename internals of Grimoire to Grimoire

wang429:
Honestly, got kind of annoyed with the length of "ForbiddenGrimoire" while adding the Grimoire exhume power. This is to make it more consistent internally.


dbjorge:
Unfortunately, we can't update the IDs of cards that have already appeared in a release build without breaking existing save files that contain the card (ie, every wanderer run), so I don't think we can do this one

PR #457, opened 2020-03-08
Implements Cull card Wasting Strike

Cartopol:
2: Deal damage equal to 10 + 2 per card in exhaust pile


src/main/java/stsjorbsmod/cards/cull/WastingStrike.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public WastingStrike() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        baseDamage = DAMAGE;
+        baseMagicNumber = DAMAGE_PER_EXHAUSTED;
+        this.tags.add(CardTags.STRIKE);
+    }
+
+    @Override
+    public int calculateBonusBaseDamage() {
```
dbjorge: Would be nice for this to calculate the total damage in an extended description

dbjorge: (see Ectoplasm/Drain Life/etc for examples)

Cartopol: The current implementation uses the same mechanism as Heavy Blade, Weight of Memory and the like. The damage value in the DESCRIPTION already shows the final calculated damage value, so having an extended description would be superfluous. Judging from similar cards, it's not particularly necessary to show the base value (before bonus calculations) on the card.

dbjorge: Sorry, you're right!

PR #459, opened 2020-03-08
Implements Cull card Polluted Strike

Cartopol:
1: [CURSE] Deal 6 damage to target once for each Curse in your hand.

Implemented similar to an an attack (so it scales with Strength etc), but with type Curse (and as such it does not trigger kunai etc.). Another note is that playing this while you have blue candle, you do lose 1 hp and exhaust this, but that seems in line with blue candle wording ("whenever you play a curse...")


src/main/java/stsjorbsmod/cards/cull/PollutedStrike.java
```diff
@@ -0,0 +1,58 @@
...
+import stsjorbsmod.JorbsMod;
+import stsjorbsmod.cards.CustomJorbsModCard;
+import stsjorbsmod.characters.Cull;
+
+public class PollutedStrike extends CustomJorbsModCard {
+    public static final String ID = JorbsMod.makeID(PollutedStrike.class);
+
+    private static final CardRarity RARITY = CardRarity.COMMON;
+    private static final CardTarget TARGET = CardTarget.ENEMY;
+    private static final CardType TYPE = CardType.CURSE;
```
dbjorge: With this being a curse, I'm not sure whether this is intended to do normal type damage or thorns type damage (like Snake Oil does, being a skill that does damage). Asked in Discord what the intent is; I think either way there should be a slight change:

* If it is intended to work like Snake Oil, it should use `damageType = damageTypeForTurn = DamageInfo.DamageType.THORNS;` and pass `damageTypeForTurn` to the `DamageInfo` ctor in `use`
* If it is intended to do normal-type damage, it should use the `stsjorbsmod:Attack\u00a0Damage` keyword in the card description to make this clear to the player

dbjorge: Response back on discord: it's correct for it to be normal/attack damage, so the implementation is fine, just recommend updating description to say `stsjorbsmod:Attack\u00a0Damage` instead of `damage`

Cartopol: I've commented on the discord that this implementation might need some change later. Having cards be dual-typed would make some differences, esspecially for a card like Pollute (turns all cards in hand into curses, giving Curse frame to attacks and powers with different art crops). 
But since this requires some complex work, I would suggest to playtest this implementation first before committing to dual-typed curses.

PR #463, opened 2020-03-11
Implement new CULL Card: Feint

Panic-Point:
I made a card :-). Learned a lot doing it and hope to contribute a lot more going forward. 
First time working on anything like this so I hope it looks good.

1: Deal 6(10) Damage. Draw 2 cards, then discard a card of your choice.


Frenoss:
Thanks for jumping in! I felt the same way when working on my first card. Lots of learning...

I only have a few small suggestions, as noted above. dbjorge's a bit occupied this week, so the review might take some more time to complete.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -217,6 +217,10 @@
     "NAME": "Feast",
     "DESCRIPTION": "Remember stsjorbsmod:Gluttony. NL Draw !M! cards. NL Exhaust."
   },
+  "stsjorbsmod:Feint": {
+    "NAME": "Feint",
+    "DESCRIPTION": "Deal !D! Damage. NL Draw 2 Cards. NL Discard 1 Card."
```
dbjorge: The terms "card" and "damage" usually aren't capitalized in card descriptions (when in doubt, follow the example of existing cards :) )

dbjorge: We prefer to use placeholders for numbers on cards rather than hardcoding them into the description strings; even if they don't ever change in-game (like these ones), numbers frequently change during later balancing passes, and using placeholders from the start means fewer changes to all the translations' strings files for simple balance changes.

Since these effects aren't block/damage, you would want to use "magic number" placeholders for these 2 effects. For example, you might do something like

`"Deal !D! damage. NL Draw !M! cards. NL Discard !stsjorbsmod:MetaMagic! card."`

...and then fill in the placeholders by adding lines to the constructor for the card implementation:

```java
baseMagicNumber = magicNumber = DRAW;
baseMetaMagicNumber = metaMagicNumber = DISCARD;
```

dbjorge:
If possible, it would be nice to mspaint a basic beta card art for this; even if it's just a stick figure on a solid colored background, it's helpful to have a unique image for each card for the purposes of making it easy for a player to distinguish the cards in their hand.

Ideally, you want a 500x380 png in the "originals" folder (easiest to copy-paste an existing one as a starting point to get the right canvas size)

We have a script to create the "generated" images automatically based on the "originals" images. Making the generated images is optional (I don't mind running the script to generate them as part of merging the commit), but if you want to do it yourself, the steps are:
* (one-time) Make sure [Node.js](https://nodejs.org/en/) is installed and on your PATH
* (one-time) Make sure ImageMagick is installed and on your PATH (on Windows, the easiest way to do this is to install [Chocolatey](https://chocolatey.org/) and then run `choco install imagemagick` from an Administrator command prompt
* (each time you make a card) From a command prompt at the root of repository, run `node scripts/generate-card-images.js Feint`

This should produce 2 images (`Feint_p.png` and `Feint.png` under a `/generated/` folder)

Panic-Point:
I will make the updates now. Some of these notes will probably apply to all my pull requests, for example the image stuff. I will update all of them with the notes here.

PR #464, opened 2020-03-11
Implements Cull cards Cleanse and Anxiety

Cartopol:
1: Discard all curses from your drawpile and heal 1hp per curse discarded. Exhaust


Cartopol:
Whoops, I accidentally continued work on my Cleanse branch when making Anxiety, so this PR contains two cards.

src/main/java/stsjorbsmod/cards/cull/Anxiety.java
```diff
@@ -0,0 +1,62 @@
...
+        super.applyPowers();
+        this.initializeDescription();
+    }
+
+
+    @Override
+    public void upgrade() {
+        if (!upgraded) {
+            upgradeMetaMagicNumber(UPGRADE_DRAW);
+            ++this.misc;
```
dbjorge: Should  this be `this.misc += UPGRADE_DRAW;`?

src/main/java/stsjorbsmod/cards/cull/Anxiety.java
```diff
@@ -0,0 +1,62 @@
...
+    public void use(AbstractPlayer p, AbstractMonster abstractMonster) {
+        addToBot(new DrawCardAction(this.misc));
+        addToBot(new GainEnergyAction(this.urMagicNumber));
+        if (this.misc > 0) {
+            this.addToBot(new IncreaseMiscAction(this.uuid, this.misc, -this.magicNumber));
+        }
+    }
+
+    public void applyPowers() {
+        this.baseMetaMagicNumber = this.misc;
```
dbjorge: I think you'll want to also override `applyLoadedMiscValue` with this line, to ensure the number will show correctly in the master deck after a save/load transition

Cartopol: I'm not a 100% sure if I did the override correctly. I'm calling super.applyLoadedMiscValue and initializeDescription, but I'm not totally sure if that's needed for anything.

PR #465, opened 2020-03-12
Implement CULL Card: Godsbane

Panic-Point:
1: Deal 8(10)% of targets max HP damage to it.

Wasn't sure how STS deals with doubles and rounding so I did it all here.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -268,6 +268,10 @@
     "DESCRIPTION": "Draw !M! card. NL This turn, your next attack is played !stsjorbsmod:MetaMagic! additional time.",
     "UPGRADE_DESCRIPTION": "Draw !M! card. NL This turn, your next attack is played !stsjorbsmod:MetaMagic! additional times."
   },
+  "stsjorbsmod:Godsbane": {
+    "NAME": "Godsbane",
+    "DESCRIPTION": "Deal !M! % of targets max HP."
```
Frenoss: I would add an apostrophe "target's" and the word "as damage" after the HP.

If you use !D!, will it show the actual damage value for a given target? That seems useful so people know the exact number without having to guess or math our implementation. Altogether that might lead to a DESCRIPTION like

Deal !M!% (!D!) of target's max HP as damage.

src/main/java/stsjorbsmod/cards/cull/Godsbane.java
```diff
@@ -0,0 +1,53 @@
...
+        damage = baseDamage = 0;
+    }
+
+    private void recalculateDamage(AbstractMonster m) {
+        damage = (int) Math.round((m.maxHealth * ((double) magicNumber / 100)));
+    }
+
+    @Override
+    public void calculateCardDamage(AbstractMonster m) {
+        super.calculateCardDamage(m);
```
Frenoss: I am actually being dumb here. What's the thought behind calling the super? I'm guessing it's in the general case that the super has some side effects beyond damage calculation?

src/main/java/stsjorbsmod/cards/cull/Godsbane.java
```diff
@@ -0,0 +1,53 @@
...
+    public static final String ID = JorbsMod.makeID(Godsbane.class);
+
+    private static final CardRarity RARITY = CardRarity.COMMON;
+    private static final CardTarget TARGET = CardTarget.ENEMY;
+    private static final CardType TYPE = CardType.ATTACK;
+    public static final CardColor COLOR = Cull.Enums.CULL_CARD_COLOR;
+
+    private static final int COST = 1;
+    private static final int DAMAGE_PERCENT = 8;
+    private static final int UPGRADE_DAMAGE_PERCENT = 2;
```
Frenoss: Seems like another case where _PLUS would help communicate how this upgrade works better than it is named now.

src/main/java/stsjorbsmod/cards/cull/Godsbane.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public Godsbane() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        magicNumber = baseMagicNumber = DAMAGE_PERCENT;
+        damage = baseDamage = 0;
+    }
+
+    private void recalculateDamage(AbstractMonster m) {
```
Frenoss: More curiosity on my part--what's your thinking around having this in a method instead of in the calculateCardDamage() override? In terms of the idea of what it's doing, it seems to me like it exactly is an override for calculateCardDamage(). Just hoping to spark discussion and learn more myself.

Panic-Point: Hey Fresnoss, Thanks for the reviews. Mainly I was attempting to emulate different cards that I saw that were similar. You're right though, this seems like it should work within the override itself. I will test it again tonight and make sure that it works.

src/main/java/stsjorbsmod/cards/cull/Godsbane.java
```diff
@@ -0,0 +1,53 @@
...
+        damage = (int) Math.round((m.maxHealth * ((double) magicNumber / 100)));
```
wang429: Please double check the expected rounding behavior in the Discord. I imagine ceiling might be more expected.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
Panic-Point: adding the !D! just shows 0 on the card. Though I imagine there is a way to have it update when you mouse over a target, similar to the update if the target is vulnerable. Not sure of how to make it do that though.

Frenoss: It was a good go. Let's call that a future feature.

src/main/java/stsjorbsmod/cards/cull/Godsbane.java
Frenoss: Oh yeah, following the existing cards is the way to go.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
Cartopol: Hey Panic-Point, you can use some features of the CustomJorbsmodCard class to calculate the damage for these kinds of purposes. Look at a card like Drain Life for how to implement it. For a card like this, you can set the baseDamage to 0, and then in method calculateBonusBaseDamage you calculate the percentage of the target's HP. This allows you to use the object's damage property within your use method, and allows you to refer to !D! in an extended description. 
This is a great feature to get used to, especially for displaying calculated values on a card.

Panic-Point: Thanks for helping find this Cartopol. AbstractDungeon.getCurrRoom().monsters.hoveredMonster seemed to do the trick for showing the actual damage that will be dealt.

src/main/java/stsjorbsmod/cards/cull/Godsbane.java
Panic-Point: Checked and STS just casts damage to int and doesn't round. Updated to reflect that.

dbjorge:
The implementation looks pretty good as-is; the only thing I noted as being off was that the extended description was "sticking" to the card after a monster became unhovered in a way that wasn't desireable. The fix for this was a little complicated to identify (though not too complicated to implement); I went ahead and pushed it into the branch here.

PR #466, opened 2020-03-12
Implement new CULL Card: Goosebumps

Panic-Point:
1: Deal 3 damage to target 3(4) times.


src/main/java/stsjorbsmod/cards/cull/Goosebumps.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private static final CardRarity RARITY = CardRarity.COMMON;
+    private static final CardTarget TARGET = CardTarget.ENEMY;
+    private static final CardType TYPE = CardType.ATTACK;
+    public static final CardColor COLOR = Cull.Enums.CULL_CARD_COLOR;
+
+    private static final int COST = 1;
+    private static final int DAMAGE = 3;
+    private static final int MULTI = 3;
+    private static final int UPGRADE_MULTI = 1;
```
Frenoss: Probably unnecessary repetition on my part after the other reviews, but I would note _PLUS in the name for this upgrade. Perhaps rename to ATTACK_COUNT and UPGRADE_PLUS_ATTACK_COUNT?

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -268,6 +268,10 @@
     "DESCRIPTION": "Draw !M! card. NL This turn, your next attack is played !stsjorbsmod:MetaMagic! additional time.",
     "UPGRADE_DESCRIPTION": "Draw !M! card. NL This turn, your next attack is played !stsjorbsmod:MetaMagic! additional times."
   },
+  "stsjorbsmod:Goosebumps": {
+    "NAME": "Goosebumps",
+    "DESCRIPTION": "Deal !D! damage to target !M! times."
```
Frenoss: I'm bad for not checking, but I think templating might let us get away with being even simpler and dropping the "to target" from the description.

Panic-Point:
From what I looked at, the isMultiDamage seems to correspond to damaging multiple enemies rather than multiple damage to the same enemy. Things like flechettes don't set it to true, but Doom and Gloom does.

Also if I try to use multiDamage for the DamageInfo it throws a type error. Happy to update if there is a better way to do this, but from what I see adding multiple attack options is typically how this is handled.

For the other updates, your wording is way better. I will make that change

Frenoss:
Ohh, that makes sense about the meaning of isMultiDamage. Sorry for sending you on that side quest. Since this was your initial foray to helping with the project, I wanted to be sure your pull requests got some eyes on them right away, even if they're my rusty ones. It's great to see the diligence you've put in!

PR #468, opened 2020-03-13
Implement new CULL Card: Melt

Panic-Point:
1: Apply burning to enemy equal to its shield. Deal 8(12) damage.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -364,6 +364,10 @@
     "NAME": "Materialize",
     "DESCRIPTION": "Gain !B! Block. NL Choose 1 of 3 stsjorbsmod:Material Components."
   },
+  "stsjorbsmod:Melt": {
+    "NAME": "Melt",
+    "DESCRIPTION": "Apply burning equal to target's block. NL Deal !D! damage."
```
dbjorge: Burning should be keyworded (`stsjorbsmod:Burning`)

dbjorge: `Block` is usually capitalized in card descriptions (I don't know why the base game capitalizes `Block` but not `damage`...)

PR #469, opened 2020-03-13
Deck of trials

wang429:
New side deck: Deck of Toils - unique cards which generally exert
New Relic: Book of Toils - at the start of an act, draw 2 random cards from the Deck of Trials and add them to your deck.

Functionally, they get added after Neow bonus chosen or between acts.


wang429:
TODO: make cards non-duplicateable. still need to figure out if they're removeable
Book of Toils relic image doesn't show as expected.

src/main/java/stsjorbsmod/JorbsMod.java
```diff
@@ -91,6 +91,10 @@
         // Use on a card that remembers a memory, which is mechanic specific to the Wanderer character.
         @SpireEnum(name = "REMEMBER_MEMORY")
         public static AbstractCard.CardTags REMEMBER_MEMORY;
+
+        // Cards in the Deck of Toils should be
+        @SpireEnum(name = "DECK_OF_TOILS")
```
dbjorge: I think it's "trials", not "toils"

wang429: I'll check in Discord. it's definitely inconsistent, Book of Toils relic description indicates "toils", bottom section indicates "trials"

wang429: trials it is!

wang429:
oh didn't notice the conflicts, I'll handle this when I've got some time.

src/main/java/stsjorbsmod/patches/LegendaryPatch.java
```diff
@@ -32,7 +32,9 @@
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import static stsjorbsmod.JorbsMod.JorbsCardTags.DECK_OF_TRIALS;
```
dbjorge: I'm confused about why the changes to `LegendaryPatch` would be necessary; I'd have expected all the deck of trials cards to just be `SPECIAL` rarity and to never appear in the first place except via the `DeckOfTrials`

src/main/java/stsjorbsmod/cards/cull/deckoftrials/DeckOfTrials.java
```diff
@@ -0,0 +1,42 @@
...
+            if (!deck.isEmpty()) {
+                AbstractCard card = deck.getTopCard();
+                deck.removeCard(card);
+                cards.add(card);
+            }
+        }
+        return cards;
+    }
+
+    public static void addDeckOfTrialsCardsToMasterDeck() {
```
dbjorge: I think this method would make more sense on the Book of Trials relic

src/main/java/stsjorbsmod/cards/cull/deckoftrials/DeckOfTrials.java
```diff
@@ -0,0 +1,42 @@
...
+        deck.addToTop(new Frostbite());
+        deck.shuffle();
+    }
+
+    public static ArrayList<AbstractCard> drawCards(int num) {
+        ArrayList<AbstractCard> cards = new ArrayList<>();
+        for (int i = 0; i < num; i++) {
+            if (!deck.isEmpty()) {
+                AbstractCard card = deck.getTopCard();
+                deck.removeCard(card);
```
dbjorge: We'll probably want to add a CustomSaveable implementation for this deck state, to avoid players being able to get duplicate trials cards after a save/load.

dbjorge: (ideally we'd want to do a specific saveable for this rather than relying on enumerating the player's master deck on load, since players can remove cards)

src/main/java/stsjorbsmod/patches/CardCrawlGamePatch.java
```diff
@@ -0,0 +1,27 @@
...
+public class CardCrawlGamePatch {
+
+    @SpirePatch(clz = CardCrawlGame.class, method = "update")
+    public static class CardCrawlGame_Update {
+        public static ExprEditor Instrument() {
+            return new ExprEditor() {
+                public void edit(MethodCall methodCall) throws CannotCompileException {
+                    if (methodCall.getClassName().equals(DungeonMapScreen.class.getName()) && methodCall.getMethodName().equals("open")) {
+                        methodCall.replace(String.format(
+                                "{ %1$s.addDeckOfTrialsCardsToMasterDeck(); $_ = $proceed($$); }",
```
dbjorge: I think I would probably lean towards making this (and the Neow event patches) slightly more generally named; maybe something like `AtStartOfActRelicSubscriber::atStartOfAct()`, and make the Book's implementation the thing responsible for associating "the start of acts" with "add deck of trials cards"

src/main/java/stsjorbsmod/patches/LegendaryPatch.java
wang429: Admittedly, I don't remember; and probably didn't have a particular reason. 🤦‍♂

src/main/java/stsjorbsmod/cards/cull/deckoftrials/DeckOfTrials.java
wang429: As it stands, it has a different issue of not reverting to the original deck state on save/load. Since the object is static, it stays in memory and on save/load after entering a new act, the deck of trials contains two fewer cards.

wang429: And because of the way the deck is instantiated, on loading the run after killing STS, the deck of trials is completely empty

src/main/java/stsjorbsmod/JorbsMod.java
```diff
@@ -96,6 +96,10 @@
         // Use on a card that remembers a memory, which is mechanic specific to the Wanderer character.
         @SpireEnum(name = "REMEMBER_MEMORY")
         public static AbstractCard.CardTags REMEMBER_MEMORY;
+
+        // Cards in the Deck of Trials should be
+        @SpireEnum(name = "DECK_OF_TRIALS")
+        public static AbstractCard.CardTags DECK_OF_TRIALS;
```
dbjorge, line 102: Don't think this serves any purpose anymore, we can omit it until it's needed

wang429:
Sorry this took so long

PR #473, opened 2020-03-14
Implement new Cull Card: Overkill

Panic-Point:
1: [Power] - Deal overkill damage (+5) to all other enemies


src/main/java/stsjorbsmod/powers/OverkillPower.java
```diff
@@ -0,0 +1,48 @@
...
+            this.description = String.format(DESCRIPTIONS[1], this.amount);
+        }
+    }
+
+
+    @Override
+    public void onAttack(DamageInfo info, int damageAmount, AbstractCreature target) {
+        if(damageAmount > target.currentHealth){
+            int overkill_damage = damageAmount - target.currentHealth;
+            addToBot(new DamageAllEnemiesAction(null,
```
dbjorge: Should probably use `this.owner` as the damage source

src/main/java/stsjorbsmod/powers/OverkillPower.java
```diff
@@ -0,0 +1,48 @@
...
+                    DamageInfo.createDamageMatrix(overkill_damage + this.amount),
+                    DamageInfo.DamageType.NORMAL, AbstractGameAction.AttackEffect.SLASH_HORIZONTAL, true));
```
dbjorge: Powers that do damage like (thorns, combust, etc) this typically do `DamageType.THORNS`, rather than `NORMAL`

src/main/java/stsjorbsmod/powers/OverkillPower.java
```diff
@@ -0,0 +1,48 @@
...
+        else {
+            this.description = String.format(DESCRIPTIONS[1], this.amount);
+        }
+    }
+
+
+    @Override
+    public void onAttack(DamageInfo info, int damageAmount, AbstractCreature target) {
+        if(damageAmount > target.currentHealth){
+            int overkill_damage = damageAmount - target.currentHealth;
```
dbjorge: We usually prefer `camelCase` rather than `snake_case` for local variable names

src/main/java/stsjorbsmod/powers/OverkillPower.java
dbjorge: (you'll want to pass `isPureDamage: true` to `createDamageMatrix` to coincide with this)

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -161,6 +161,13 @@
       "You cannot remember other memories until you have #yClarity of your current memory."
     ]
   },
+  "stsjorbsmod:OverkillPower": {
+    "NAME": "Overkill Power",
```
dbjorge: The `Power` in the name is implied; the name here can just be `Overkill`

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -414,6 +414,11 @@
     "DESCRIPTION": "Gain !M! gold. NL Remember stsjorbsmod:Charity. NL Exhaust.",
     "UPGRADE_DESCRIPTION": "Gain !M! gold. NL Remember stsjorbsmod:Charity. NL Gain stsjorbsmod:Clarity of current memory. NL Exhaust."
   },
+  "stsjorbsmod:Overkill": {
+    "NAME": "Overkill",
+    "DESCRIPTION": "Deal damage equal to overkill damage to all other enemies.",
```
dbjorge: The way this is worded made me think it was going to be dealing damage immediately; maybe instead, `Whenever you kill an enemy, deal any overkill damage to ALL enemies.`?

dbjorge: (similar feedback for power description)

src/main/java/stsjorbsmod/powers/OverkillPower.java
```diff
@@ -0,0 +1,48 @@
...
+            this.description = String.format(DESCRIPTIONS[0]);
+        }
+        else {
+            this.description = String.format(DESCRIPTIONS[1], this.amount);
+        }
+    }
+
+
+    @Override
+    public void onAttack(DamageInfo info, int damageAmount, AbstractCreature target) {
```
dbjorge: Using this hook makes me a little nervous because in theory a monster power with an `onAttacked` hook could change the damage amount between this hook and the monster's death, but in practice I don't see any non-deprecated powers that use `onAttacked` to change damage amount, so I think this is probably fine

PR #474, opened 2020-03-14
Cull mechanic exert

Cartopol:
In the current implementation, if you want a card to be exerted when used, you have to set the EntombedField of the masterDeckEquivalent to true, as well as set the new CustomJorbsModCard.exert property to true. Cards with this property are checked at the start of the act, setting their Entombed property to false. This is a little clumsy in the card implementation, so maybe there's a smarter way.
Currently there is no way to distinguish an exerted card from an exhausted card in the exhaust pile. Also, a card has no 'exerted' property to hook onto, it is all managed using the entombed mechanics. That might be a welcome addition.

This PR includes the card 'Frustration', made for testing the exert mechanics. It is fully functional and can be reviewed on its own merits.


src/main/java/stsjorbsmod/cards/CustomJorbsModCard.java
```diff
@@ -38,6 +37,9 @@
     public boolean upgradedMetaMagicNumber = false;
     public boolean isMetaMagicNumberModified = false;
 
+    public boolean exert = false;
```
dbjorge: I would recommend using a SpireField on AbstractCard rather than a field on CustomJorbsModCard; the latter will cause problems with cards like `Shrapnel Bloom` that can Exert arbitrary other cards (including base game colorless cards)

src/main/java/stsjorbsmod/patches/AbstractCardMakeStatEquivalentCopyPatch.java
```diff
@@ -31,6 +31,11 @@ public static AbstractCard Postfix(AbstractCard __result, AbstractCard __this) {
             card.baseMetaMagicNumber = source.baseMetaMagicNumber;
             card.isMetaMagicNumberModified = source.isMetaMagicNumberModified;
             card.upgradedMetaMagicNumber = source.upgradedMetaMagicNumber;
+            card.exert = source.exert;
+
+            boolean entombed = EntombedField.entombed.get(source);
```
dbjorge: This should go outside the `if` block (like `WrathField`)

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Keyword-Strings.json
```diff
@@ -48,6 +48,14 @@
     "NAMES": ["entombed"],
     "DESCRIPTION": "This card starts each combat in your Exhaust pile. It will be Exhumed from your Exhaust pile into your hand each time its Exhume condition is met."
   },
+  {
+    "NAMES": ["exert"],
+    "DESCRIPTION": "Exhaust this card when played. It will be entombed until the start of next Act."
```
dbjorge: "Entombed" should be capitalized and `#y`'d

src/main/java/stsjorbsmod/cards/cull/Frustration.java
```diff
@@ -0,0 +1,61 @@
...
+    @Override
+    public void onRetained() {
+        this.addToBot(new DamageAllEnemiesAction(null, DamageInfo.createDamageMatrix(magicNumber, true), DamageInfo.DamageType.THORNS, AbstractGameAction.AttackEffect.FIRE));
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster abstractMonster) {
+        this.addToBot(new DamageAllEnemiesAction(p, this.multiDamage, this.damageTypeForTurn, AbstractGameAction.AttackEffect.SLASH_DIAGONAL));
+
+        AbstractCard masterCard = StSLib.getMasterDeckEquivalent(this);
```
dbjorge: I would probably lean towards trying to treat this more like how `this.exhaust` works, where if `this.exert` is true, the card exhausts and sets its master deck equivalent to "exerted" automatically when played, rather than requiring each individual card to set all of that explicitly.

I think it'd be okay to do a similar compromise option to what `EphemeralField` does, where setting `ExertField` to true on a card implicitly also sets `card.exhaust` to true.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -263,6 +263,10 @@
     "NAME": "Frostbite",
     "DESCRIPTION": "Deal !D! damage. NL If this kills an enemy, stsjorbsmod:Consume this."
   },
+  "stsjorbsmod:Frustration": {
+    "NAME": "Frustration",
+    "DESCRIPTION": "Retain. NL Deal !D! damage to ALL enemies. NL Whenever this is retained, deal !M! damage to ALL enemies. NL stsjorbsmod:Exert."
```
dbjorge: I think the wording that you've done here is closer to how the base game would word it than the version in the spreadsheet, but I think there are planned to be enough "retain / exert" sort of cards and that they read so much better with the spreadsheet version that I'm okay establishing the convention from the spreadsheet for them. `Retain: Deal !M! damage to ALL enemies. NL stsjorbsmod:Exert: Deal !D! damage to ALL enemies.`

(assuming the `stsjorbsmod:Exert:` with the colon on the end actually works in practice; I'm suspicious)

src/main/java/stsjorbsmod/cards/CustomJorbsModCard.java
```diff
@@ -238,4 +240,11 @@ public void triggerOnGlowCheck() {
         }
         return null;
     }
+
+    @Override
+    public void receiveStartAct() {
+        if (this.exert) {
```
dbjorge: I don't think this is going to do the right thing with `Shrapnel Bloom`, where it wants to make cards exert as a one-time thing without necessarily being permanently exert-on-play. I think probably it would make more sense to have separate fields for `exerted` vs `selfExert`/`exert`

src/main/java/stsjorbsmod/cards/cull/Frustration.java
Cartopol: Agreed. Where do I actually patch in the stuff that causes the card's ExertedField to get set? Surely it's not just a Postfix on AbstractCard::Use, right? I'm trying to find the spot where a card with this.exhaust triggers a move to the exhaust pile, but I can't seem to find it.

Cartopol: Ah, found it in UseCardAction (makes sense, I just had never seen this being called). I'll see if I can figure out how to patch it. But that's not my forte, so any help is much appreciated :)

Cartopol: Alright, I think I figured it out. Hope it's not too crude a solution.

src/main/java/stsjorbsmod/JorbsMod.java
```diff
@@ -418,4 +419,13 @@ public void receivePowersModified() {
             }
         }
     }
+
+    @Override
+    public void receiveStartAct() {
+        for (AbstractCard c : AbstractDungeon.player.masterDeck.group) {
+            if (c instanceof StartActSubscriber) {
+                ((StartActSubscriber)c).receiveStartAct();
```
dbjorge: Repeating a comment from #476 that also applies here:

> I think forwarding the `StartGameSubscriber` and `StartActSubscriber` notifications is a perfectly reasonable design for Equinox. However, I would not recommend forwarding this notification to *every* master deck card that implements  the BaseMod subscribers like this, because it would be possible in theory for other mods to exist that have cards of their own which implement the BaseMod subscribers and have their own separate mechanisms for forwarding the same events, and we don't want to cause such a mod to receive 2 copies of those events.
>
> I think having `JorbsMod` implement BaseMod's `receiveStartAct` and `receiveStartGame` is good, but I'd suggest not using the BaseMod interfaces for the card; instead, I'd recommend adding equivalent methods on `CustomJorbsModCard` (with no-op default behavior), and only have the `JorbsMod` hooks invoke the `CustomJorbsModCard` hooks.

src/main/java/stsjorbsmod/cards/cull/Frustration.java
```diff
@@ -0,0 +1,56 @@
...
+    private static final int UPGRADE_DAMAGE_ON_RETAIN = 1;
+
+
+    public Frustration() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        this.selfRetain = true;
+        this.isMultiDamage = true;
+        this.baseDamage = DAMAGE;
+        this.baseMagicNumber = DAMAGE_ON_RETAIN;
+        this.exhaust = true;
```
dbjorge: `this.exhaust` is implied by `selfExert` now

src/main/java/stsjorbsmod/patches/ExertedField.java
```diff
@@ -0,0 +1,14 @@
...
+import com.evacipated.cardcrawl.modthespire.lib.SpirePatch;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+
+@SpirePatch(
+        clz = AbstractCard.class,
+        method = SpirePatch.CLASS
+)
+public class ExertedField {
+    @SuppressWarnings("unchecked")
+    public static SpireField<Boolean> exerted = new SpireField(() -> false);
```
dbjorge: We'll probably need to add some bit of handling to make sure this state gets persisted across save/load (if we aren't already doing that for entombed, too, we probably should)

src/main/java/stsjorbsmod/cards/CustomJorbsModCard.java
```diff
@@ -238,4 +237,11 @@ public void triggerOnGlowCheck() {
         }
         return null;
     }
+
+    @Override
+    public void receiveStartAct() {
```
dbjorge: For a case like this where there is base class functionality we want to always happen and also we want to make the same hook something that cards like Equinox can override, I'd recommend a pattern like this:

```java
// This is what JorbsMod::receiveStartAct calls; it is final, so individual cards can't override it
public final void receiveStartAct() {
    // <base class exerted functionality>
    this.atStartOfAct();
}
// Individual cards override this (and don't have to remember to call super.sameMethod() because of the pattern)
protected void atStartOfAct() { }
```

src/main/java/stsjorbsmod/cards/cull/Frustration.java
dbjorge: LGTM!

src/main/java/stsjorbsmod/cards/CustomJorbsModCard.java
dbjorge: Oh, oops, except that we shouldn't be using `CustomJorbsModCard` for this particular bit of functionality, because it needs to apply to base game cards that get exerted as well

src/main/java/stsjorbsmod/JorbsMod.java
dbjorge: Actually, I just remembered that this particular start-of-act behavior also needs to apply to base game cards that get exerted, so this probably should just inline the "reset exerted" behavior

Cartopol: Thanks for the review! There's a couple things I don't quite understand regarding the implementation of your requests. Maybe there's just something I'm missing from how hooks can be implemented.
- Since we can set base cards to be Exerted, we still want to call the receiveStartAct method for each card in the MasterDeck, right? And not just for instanceof CustomJorbsModCard? But if I check card instanceof StartActSubscriber, the base cards won't receive this event, do they?
- Where do I place the logic for setting ExertedField to false at receiveStartAct? This should be general game logic, so putting it in a CustomJorbsModCard method would not work right for base cards? 

Hope you don't mind providing a little more guidance on this one!

src/main/java/stsjorbsmod/patches/ExertedField.java
Cartopol: Would it break saves if the CardData object (and corresponding json) gets Entombed and Exerted values?

src/main/java/stsjorbsmod/JorbsMod.java
Cartopol: Ah, for some reason I didn't see your last reply. I'll push my current version. It does the reset of Exerted in Jorbsmod::receiveStartAct for all cards in MasterDeck. It also calls receiveStartAct for each card that's an instanceof StartActSubscriber. And lastly I've added your suggested pattern in CustomJorbsModCard. Hope I'm doing all that stuff correctly this time!

src/main/java/stsjorbsmod/patches/ExertedField.java
dbjorge: That looks good to me!

src/main/java/stsjorbsmod/JorbsMod.java
dbjorge: The exert reset looks good in the current iteration, but this part:

> It also calls receiveStartAct for each card that's an instanceof StartActSubscriber.

...is the thing that still concerns me.

Consider what happens if *a second mod* that the user had installed were to try to implement similar functionality in the same way; two different mods' entry point classes would *both* be invoking the `receiveStartAct` hook on each card in the master deck that implements BaseMod's `StartActSubscriber`, so each such card would end up running its hooks multiple times.

We can avoid such an issue by just avoiding the use of the BaseMod `StartActSubscriber` interface on our cards or our `CustomJorbsModCard` base class; you can still have a method by the same name on the `CustomJorbsModCard` base class, just don't implement the BaseMod interface anywhere except `JorbsMod`, and have `JorbsMod`'s `receiveStartAct` hook look for cards that implement `CustomJorbsModCard` rather than cards that implement BaseMod's `StartActSubscriber`.

Cartopol: Oh, of course! I was still thinking we needed send this event to all cards to get the Exert reset stuff to work, but of course that's not the case anymore. 
Just pushed a correction. Hopefully that's the last iteration.

PR #475, opened 2020-03-15
Cull card changes 20200402

Cartopol:
- Cull changed to: Deal 12 damage twice. Exhaust.
- Cull upgrade changed to: gains an additional wrath stack. Wording of this effect on the card could use work.
- Haunt changed to: Ephemeral(Exhaust). Gain 1 intangible and draw 1 fewer card next turn.
- Abjure text changed to represent that you will have exactly 1 Spirit Shield.
- Miracle: added Exhaust
- Siphon: changed damage, heal amount, upgrade heal amount and rarity. Removed from the starting loadout, since it is now a common.
- Ectoplasm: changed gold gained from 10 to 5.
- Wail: added Exhaust. Changed rarity from basic to uncommon, and removed from the starting loadout.
- Apparate: changed upgrade damage increase from 3 to 4. Changed rarity from basic to uncommon, and removed from the starting loadout.
- Toil: changed to work only on non-minion enemies, using the Fatal keyword.
- Wasting Essance: changed damage per curse drawn and upgrade damage increase.
- Strife: only counts curses in discard pile. Upgrade adds 2 curses instead of 2 strength per curse. Added Exhaust.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -127,7 +127,8 @@
   },
   "stsjorbsmod:CULLCard": {
     "NAME": "CULL",
-    "DESCRIPTION": "Permanently decrease this card's damage by !stsjorbsmod:UrMagic! each time it is drawn. NL Deal !D! damage twice. NL If Fatal, permanently increase this card's damage by !M!. NL Exhaust."
+    "DESCRIPTION": "Deal !D! damage twice. NL Exhaust.",
+    "UPGRADE_DESCRIPTION": "Deal !D! damage twice. NL stsjorbsmod:Wrath is applied twice. NL Exhaust."
```
dbjorge: Maybe repeating "if Fatal" from Wrath would make the wording more clear that it's doubling the amount of new wrath stacks gained in the future, rather than doubling the effects of any previously gained stacks? eg, `If Fatal, gain twice the normal stsjorbsmod:Wrath.`?

src/main/java/stsjorbsmod/cards/cull/CULLCard.java
```diff
@@ -28,30 +22,19 @@
...
-    private static final int DAMAGE = 10;
-    private static final int DRAW_MINUS_DAMAGE = 1;
-    private static final int KILL_PLUS_DAMAGE = 3;
-    private static final int UPGRADE_PLUS_KILL_PLUS_DAMAGE = 1;
+    private static final int DAMAGE = 12;
 
     public CULLCard() {
         super(ID, COST, TYPE, COLOR, RARITY, TARGET);
-        misc = baseDamage = DAMAGE;
+        this.baseDamage = DAMAGE;
```
dbjorge: This will want a `WrathMemory::usesMiscToTrackPermanentBaseDamage` update

src/main/java/stsjorbsmod/cards/cull/Siphon.java
```diff
@@ -12,15 +12,17 @@
...
     public static final CardColor COLOR = Cull.Enums.CULL_CARD_COLOR;
 
     private static final int COST = 1;
-    private static final int DAMAGE = 10;
-    private static final int UPGRADE_DAMAGE = 4;
-    private static final int HEAL_ON_FATAL = 5;
+    private static final int DAMAGE = 9;
+    private static final int UPGRADE_DAMAGE = 3;
+    private static final int HEAL_ON_FATAL = 4;
+    private static final int UPGRADE_PLUS_HEAL_ON_FATAL = 4;
```
dbjorge: This is intended to be 1, I think

src/main/java/stsjorbsmod/cards/cull/WastingEssence.java
```diff
@@ -22,8 +22,8 @@
 
     private static final int COST = 3;
     private static final int NUMBER_OF_CURSES = 5;
-    private static final int DAMAGE_PER_CURSE = 20;
-    private static final int UPGRADE_DAMAGE_PER_CURSE = 10;
+    private static final int DAMAGE_PER_CURSE = 12;
+    private static final int UPGRADE_DAMAGE_PER_CURSE = 4;
 
     public WastingEssence() {
```
dbjorge: In addition to the damage nerf, this was renamed to `Wasting Form` (we avoid renaming classes/IDs for Wanderer cards since it's a "released" character to avoid breaking save files, but I think it's alright to do it for CULL cards at this point in its development)

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
Cartopol: I thought about that, but Wrath is also applied if another non-attack source deals the final damage. So Fatal, or any kind if "if this kills...." is unfortunately incorrect. This is a tricky one to get right without it getting very wordy!

dbjorge: Oh, that's a very good point, I had forgotten about that effect of Wrath. Let's try it as-is and iterate if we hear feedback about folks getting confused.

PR #476, opened 2020-03-16
Implements Cull card Equinox

Cartopol:
[cost = current act]: Deal damage equal to current floor number. Exhaust.

I couldn't think of an elegant way to establish the dynamic cost of this card, so I had to implement 2 new basemod subscribers, one that updates the cost whenever you change acts, and one that determines the correct cost when loading a game. If there's an easier way to approach this, please let me know!


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -194,6 +194,14 @@
     "DESCRIPTION": "Apply stsjorbsmod:Slow to ALL enemies. NL Add !M! stsjorbsmod:Web into your hand.",
     "UPGRADE_DESCRIPTION": "Apply stsjorbsmod:Slow to ALL enemies. NL Add !M! stsjorbsmod:Webs into your hand."
   },
+  "stsjorbsmod:Equinox": {
+    "NAME": "Equinox",
+    "DESCRIPTION": "Deal damage equal to the current Floor number. NL The cost of this card is equal to the current Act number. NL Exhaust.",
+    "UPGRADE_DESCRIPTION": "Deal damage equal to the current Floor number. NL The cost of this card is equal to the current Act number - !M!. NL Exhaust.",
+    "EXTENDED_DESCRIPTION": [
+      " NL (Deal !D! damage.)"
```
dbjorge: The base game uses "(Deals !D! damage.)" for these parentheticals; I don't know why the base game uses inconsistent tense for them ("deals") vs other cases of the same word ("deal"), but we should be consistent

src/main/java/stsjorbsmod/JorbsMod.java
```diff
@@ -418,4 +420,22 @@ public void receivePowersModified() {
             }
         }
     }
+
+    @Override
+    public void receiveStartAct() {
+        for (AbstractCard c : AbstractDungeon.player.masterDeck.group) {
+            if (c instanceof StartActSubscriber) {
+                ((StartActSubscriber)c).receiveStartAct();
```
dbjorge: I think forwarding the `StartGameSubscriber` and `StartActSubscriber` notifications is a perfectly reasonable design for Equinox. However, I would not recommend forwarding this notification to *every* master deck card that implements  the BaseMod subscribers like this, because it would be possible in theory for other mods to exist that have cards of their own which implement the BaseMod subscribers and have their own separate mechanisms for forwarding the same events, and we don't want to cause such a mod to receive 2 copies of those events.

I think having `JorbsMod` implement BaseMod's `receiveStartAct` and `receiveStartGame` is good, but I'd suggest not using the BaseMod interfaces for the card; instead, I'd recommend adding equivalent methods on `CustomJorbsModCard` (with no-op default behavior), and only have the `JorbsMod` hooks invoke the `CustomJorbsModCard` hooks.

PR #477, opened 2020-03-16
Cull Relic: Bottled Burden

Skyl3lazer:
Added the Bottled Burden relic for Cull. 

Bottled card starts combat in your Exhaust pile, and Exhumes on turn 3.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -92,7 +92,12 @@
     "DESCRIPTIONS": [
       "At the start of your next turn, you will Exhume some cards."
     ]
-  },
+  },"stsjorbsmod:ExhumeOnTurnXPower": {
```
dbjorge: Formatting/indentation is a little off

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -92,7 +92,12 @@
...
+  "NAME": "Exhume",
+  "DESCRIPTIONS": [
+    "At the start of turn 3, you will Exhume a card."
```
dbjorge: Would probably be better to use placeholders here (ideally for both the turn number and the card name), something like `At the start of turn #b%1$s, Exhume %2$s."

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Relic-Strings.json
```diff
@@ -6,6 +6,16 @@
       "Reduces stsjorbsmod:Burning falloff rate to 10% per turn."
     ]
   },
+  "stsjorbsmod:BottledBurdenRelic": {
+    "NAME": "Bottled Burden",
+    "FLAVOR": "I carry this with me, always",
+    "DESCRIPTIONS": [
+      "Upon pickup, choose a card that does not #yExert. Start each combat with that card #yEntombed. #yExhume it on turn 3.",
+      "Choose a card for ",
+      "At the start of turn 3 ",
```
dbjorge: Would be a little more localization friendly for this to be one string with a placeholder rather than 2 strings appended together (ie, `At the start of turn %1$s, Exhume %2$s.`) - the existing strings aren't perfect about this, but we try to be better with newer strings

src/main/java/stsjorbsmod/relics/BottledBurdenRelic.java
```diff
@@ -0,0 +1,139 @@
...
+        this.tips.clear();
+        this.tips.add(new PowerTip(this.name, this.description));
+        this.initializeTips();
+    }
+
+    @Override
+    public void atBattleStart() {
+        AbstractPower exhumeOnTurnXPower = new ExhumeOnTurnXPower(AbstractDungeon.player, AbstractDungeon.player.masterDeck.getSpecificCard(this.card), EXHUME_TURN);
+        //addToTop(new ExhaustSpecificCardAction(this.card, AbstractDungeon.player.drawPile));
+        addToTop(new ApplyPowerAction(AbstractDungeon.player, AbstractDungeon.player, exhumeOnTurnXPower));
```
dbjorge: I would recommend having the relic itself implement an `atTurnStart` rather than delegating that to an `InvisiblePower`, and omit the new `InvisiblePower` altogether; in general, we try to avoid `InvisiblePower` because they tend to interact with effects that target "buffs" or "debuffs" (eg, the new Amnesia curse, or similar effects from other mods) in ways that are confusing to the player

src/main/java/stsjorbsmod/relics/BottledBurdenRelic.java
```diff
@@ -0,0 +1,139 @@
...
+                setDescriptionAfterLoading();
+            }
+        }
+    }
+
+    @Override
+    public void onEquip() {
+        // follow same behavior as base game bottle relics.
+        CardGroup nonExertCards = CardUtils.getCardsForBottling(AbstractDungeon.player.masterDeck.getPurgeableCards());
+        nonExertCards.group.removeIf(c -> c.hasTag(HAS_EXERT));
```
dbjorge: It would probably be good to add a `canSpawn` override that returns false if there are no candidate cards in the master deck

src/main/java/stsjorbsmod/patches/AbstractCardMakeStatEquivalentCopyPatch.java
```diff
@@ -31,8 +31,13 @@ public static AbstractCard Postfix(AbstractCard __result, AbstractCard __this) {
             card.baseMetaMagicNumber = source.baseMetaMagicNumber;
             card.isMetaMagicNumberModified = source.isMetaMagicNumberModified;
             card.upgradedMetaMagicNumber = source.upgradedMetaMagicNumber;
+            boolean entombed = EntombedField.entombed.get(source);
```
dbjorge: Would be a little less redundant to do this regardless of the `if` block (similar to the `WrathField` at the top of this method)

src/main/java/stsjorbsmod/JorbsMod.java
```diff
@@ -91,6 +91,9 @@
         // Use on a card that remembers a memory, which is mechanic specific to the Wanderer character.
         @SpireEnum(name = "REMEMBER_MEMORY")
         public static AbstractCard.CardTags REMEMBER_MEMORY;
+
+        // Use on a card which Exerts the Culler
+        public static AbstractCard.CardTags HAS_EXERT;
```
dbjorge: #474 is probably going to end up using a `SpireField` rather than a tag for exert; I think I would probably recommend just leaving out the exert check from here for now, and if this PR ends up going in before #474, file an issue or a comment on #474 to track adding the exert check once exert is done

src/main/java/stsjorbsmod/patches/SelfExhumeFields.java
```diff
@@ -13,4 +13,5 @@
     public static SpireField<Boolean> selfExhumeOnSnap = new SpireField(() -> false);
     @SuppressWarnings("unchecked")
     public static SpireField<Boolean> selfExhumeOnKill = new SpireField(() -> false);
+    public static SpireField<Boolean> selfExhumeOnTurnX = new SpireField(() -> false);
```
dbjorge: This is probably unnecessary; `BottledBurden` can just exhume the specific card it's associated with, rather than tracking the same state in 2 places.

dbjorge (dismissed):
Thanks for the contribution! Left a few comments inline

src/main/java/stsjorbsmod/relics/BottledBurdenRelic.java
```diff
@@ -0,0 +1,153 @@
...
+        if (!this.cardSelected && !AbstractDungeon.gridSelectScreen.selectedCards.isEmpty()) {
+            this.cardSelected = true;
+            this.card = AbstractDungeon.gridSelectScreen.selectedCards.get(0);
+            inBottledBurden.set(card, true);
+            AbstractDungeon.getCurrRoom().phase = AbstractRoom.RoomPhase.COMPLETE;
+            AbstractDungeon.gridSelectScreen.selectedCards.clear();
+            this.description = String.format(DESCRIPTIONS[2], EXHUME_TURN, card.name);
+            this.tips.clear();
+            this.tips.add(new PowerTip(this.name, this.description));
+            this.initializeTips();
```
wang429, line 101: This appears to be exactly `setDescriptionAfterLoading()`. Can we reuse that?

src/main/java/stsjorbsmod/patches/BottledBurdenPatch.java
```diff
@@ -0,0 +1,101 @@
...
+    public static class AbstractCard_InBottleTornado_Locator extends SpireInsertLocator {
+        @Override
+        public int[] Locate(CtBehavior ctBehavior) throws Exception {
+            Matcher finalMatcher = new Matcher.FieldAccessMatcher(AbstractCard.class, "inBottleTornado");
+            return LineFinder.findInOrder(ctBehavior, finalMatcher);
+        }
+    }
+
+    // The Fountain of Curse Removal event removes curse cards from the deck. It selects them directly from the
+    // master deck. We don't let it remove any bottled curses.
```
wang429, line 85: Should the question mark floor spawner also remove this event if the player only has a bottled curse?

src/main/java/stsjorbsmod/relics/BottledBurdenRelic.java
```diff
@@ -0,0 +1,153 @@
...
+        if (nonExertCards.size() > 0) {
+            this.cardSelected = false;
+            if (AbstractDungeon.isScreenUp) {
+                AbstractDungeon.dynamicBanner.hide();
+                AbstractDungeon.overlayMenu.cancelButton.hide();
+                AbstractDungeon.previousScreen = AbstractDungeon.screen;
+            }
+
+            AbstractDungeon.getCurrRoom().phase = AbstractRoom.RoomPhase.INCOMPLETE;
+            AbstractDungeon.gridSelectScreen.open(nonExertCards, 1, DESCRIPTIONS[1] + this.name + LocalizedStrings.PERIOD, false, false, false, false);
```
wang429, line 76: For the string from DESCRIPTIONS, I wonder if it's better to let this be formatted instead of concatenated.

wang429 (changes requested):
Thanks for the contribution! I left a few comments in line.

src/main/java/stsjorbsmod/relics/BottledBurdenRelic.java
```diff
@@ -0,0 +1,153 @@
...
+            this.flash();
+            addToBot(new ExhumeCardsAction(card));
+            this.counter = -1;
+            this.grayscale = true;
+        }
+    }
+
+    @Override
+    public boolean canSpawn() {
+        Iterator playerCards = AbstractDungeon.player.masterDeck.group.iterator();
```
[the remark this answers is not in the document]
acourrau: Yup. Will do.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Relic-Strings.json
```diff
@@ -15,6 +15,15 @@
       "Cards added: "
     ]
   },
+  "stsjorbsmod:BottledBurdenRelic": {
+    "NAME": "Bottled Burden",
+    "FLAVOR": "I carry this with me, always",
+    "DESCRIPTIONS": [
+      "Upon pickup, choose a card that does not #yExert. Start each combat with that card #yEntombed [REMOVE_SPACE]. #yExhume it on turn %1$s.",
+      "Choose a card for ",
+      "At the start of turn #b%1$s [REMOVE_SPACE], Exhume #y%2$s [REMOVE_SPACE]."
```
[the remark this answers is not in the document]
acourrau: I just learned about this a week ago or so doing the non-legendary wording. I have it here so that the comma and period are not highlighted. I suppose we can just leave it out for consistency's sake.

src/main/java/stsjorbsmod/relics/BottledBurdenRelic.java
acourrau: Nice find. Thank you.

src/main/java/stsjorbsmod/relics/BottledBurdenRelic.java
acourrau: I followed the pattern for Bottled Memory, but I'm down to change both. I didn't want to mess with that existing pattern if there was a specific reason we did it that way. Can't think of anything.

src/main/java/stsjorbsmod/patches/BottledBurdenPatch.java
acourrau: Yeah probably. I'll patch it out

wang429 (approved):
Thanks for taking this PR to the end. I'm approving this. Though I am wondering about the exhumeonturnx images.

acourrau:
> Thanks for taking this PR to the end. I'm approving this. Though I am wondering about the exhumeonturnx images.

I'll remove those files and test. Good find, definitely should remove.

PR #478, opened 2020-03-16
Cull Card: Discordance

Skyl3lazer:
* New Card: Discordance (Cull, Uncommon Attack) - 1: Deal 5 (7) damage for each card type in your hand.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -155,6 +155,12 @@
     "NAME": "Determination",
     "DESCRIPTION": "Remember stsjorbsmod:Pride. NL At the end of this turn, stsjorbsmod:Snap. NL Exhaust.",
     "UPGRADE_DESCRIPTION": "Remember stsjorbsmod:Pride. NL Exhaust."
+  },"stsjorbsmod:Discordance": {
```
dbjorge: Formatting/indentation is a little inconsistent here (the localization files need to be understandable by non-programmers, so keeping the formatting consistent in them is especially important :) )

src/main/java/stsjorbsmod/cards/cull/Discordance.java
```diff
@@ -0,0 +1,68 @@
...
+    public static final CardColor COLOR = Cull.Enums.CULL_CARD_COLOR;
+
+    private static final int COST = 1;
+
+
+
+    public Discordance() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        this.baseDamage = 0;
+        this.baseMagicNumber = magicNumber = 5;
```
dbjorge: Minor suggestion: we've found it leads to fewer mistakes to use descriptively named constants for naming the "magicNumber" values; here, I'd probably add something like this under the `COST` definition:

```java
private static final int DAMAGE_PER_CARD_TYPE = 5;
private static final int UPGRADE_PLUS_DAMAGE_PER_CARD_TYPE = 2;
```

...and use those in place of the numbers in the ctor/upgrade methods.

src/main/java/stsjorbsmod/cards/cull/Discordance.java
```diff
@@ -0,0 +1,68 @@
+package stsjorbsmod.cards.cull;
```
dbjorge: It would be nice to give this an mspaint beta art that is distinct from `Drain Life`'s; the art quality doesn't matter at all (a stick figure drawing is fine), but it's helpful for the player to be able to distinguish cards at a glance

src/main/java/stsjorbsmod/cards/cull/Discordance.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public Discordance() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        this.baseDamage = 0;
+        this.baseMagicNumber = magicNumber = 5;
+    }
+
+    @Override
+    public int calculateBonusBaseDamage() {
+        Iterator var2 = AbstractDungeon.player.hand.group.iterator();
```
dbjorge: You can use a `Set` and a foreach loop in place of an Iterator to simplify this a bit:

```java
        HashSet<CardType> seen = new HashSet<>();
        for (AbstractCard c : AbstractDungeon.player.hand.group) {
            seen.add(c.type);
        }
```

Alternately, you can use `Stream::distinct`:

```java
int distinctTypes = (int) AbstractDungeon.player.hand.group.stream().map(c -> c.type).distinct().count();
```

dbjorge: It isn't clear to me from the spreadsheet's text whether this should count itself being an attack for the purposes of "play next card twice" effects (eg, Shrapnel Bloom) and/or "play card from a non-hand position" (eg, Channel); as-is, it will count itself in simple "play from hand" cases, but won't count itself in either of those 2 special cases. I asked in #cull-contributing on Discord what the intended behavior should be; we might need to adjust this to explicitly add `this.type` to the set of types considered depending on the answer.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+* New Card: Discordance (Cull, Uncommon Attack) - 1: Deal 5 (7) damage for each card type in your hand.
```
dbjorge: Thanks for including the changelog line!

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -155,6 +155,12 @@
...
+  "NAME": "Discordance",
+  "DESCRIPTION": "Deal damage equal to the number of card types in your hand times !M!.",
```
dbjorge: I worry that the term "card type" might not be intuitively clear to a player; I can see it being misinterpreted as "different card names". We should maybe make a `Card\u00a0Type` keyword that clarifies "Card Type means Attack, Skill, Power, Curse, or Status" or something

src/main/java/stsjorbsmod/cards/cull/Discordance.java
dbjorge: Per discord discussion, let's go with consistency with how Second Wind and Fiend Fire don't consider themselves as being "in-hand" while they're being played, even in the basic case; ie, explicitly skip `this` from consideration as part of the `seen` list

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
Skyl3lazer: Added Card Type/s tooltip

src/main/java/stsjorbsmod/cards/cull/Discordance.java
Skyl3lazer: Updated behavior to skip self

Skyl3lazer: Updated to match suggestion

src/main/java/stsjorbsmod/cards/cull/Discordance.java
Skyl3lazer: Created beta art for "Discord Dance"

src/main/java/stsjorbsmod/cards/cull/Discordance.java
Skyl3lazer: Updated per suggestion

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
Skyl3lazer: Fixed

PR #482, opened 2020-03-22
Implement new CULL Card: Toll

Panic-Point:
deal 3(4) x the sum of your relic counters.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -619,6 +619,13 @@
     "NAME": "Toil",
     "DESCRIPTION": "Deal !D! damage. NL If this kills an enemy, permanently upgrade a random card. NL Exhaust."
   },
+  "stsjorbsmod:Toll": {
+    "NAME": "Toll",
+    "DESCRIPTION": "Deal damage equal to !M! times the sum of your relic counters.",
```
dbjorge: Probably simpler for the unupgraded version to just be `"Deal damage equal to the sum of your relic counters"` and leave the `!M! times` for the upgraded description; I appreciate the attempt to be more localization friendly, but player friendliness is more important.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -619,6 +619,13 @@
...
+    "EXTENDED_DESCRIPTION": [
+      " NL (Deal !D! Damage)"
```
dbjorge: `" NL (Deals !D! damage.)"` is the usual wording/capitalization

src/main/java/stsjorbsmod/cards/cull/Toll.java
```diff
@@ -0,0 +1,63 @@
...
+            if (relic.counter >= 0) {
+                bonusDamage += relic.counter;
+            }
+        }
+        return bonusDamage*magicNumber;
+    }
+
+    @Override
+    public String getRawDynamicDescriptionSuffix() {
+        if (damage > 0) {
```
dbjorge: I think for this one it would make more sense to always return `EXTENDED_DESCRIPTION[0]`; for this card, if the calculated damage is 0, it is actually the case that it would be dealing 0 damage, it's not just a case of it being invalid to calculate the damage

src/main/java/stsjorbsmod/cards/cull/Toll.java
```diff
@@ -0,0 +1,63 @@
...
+public class Toll extends CustomJorbsModCard {
+    public static final String ID = JorbsMod.makeID(Toll.class);
+
+    private static final CardRarity RARITY = CardRarity.UNCOMMON;
+    private static final CardTarget TARGET = CardTarget.ENEMY;
+    private static final CardType TYPE = CardType.ATTACK;
+    public static final CardColor COLOR = Cull.Enums.CULL_CARD_COLOR;
+
+    private static final int COST = 1;
+    private static final int UPGRADE_PLUS_MAGIC_NUMBER = 1;
```
dbjorge: It would be good to give the constants more descriptive names than "magic number"; maybe `DAMAGE_MULTIPLIER` and `UPGRADE_PLUS_DAMAGE_MULTIPLIER`?

PR #484, opened 2020-03-29
Implement API to communicate with Twitch extension

avolny:
Implements API that allows displaying PowerTips on Twitch for streamers using SlayTheRelics Twitch extension

- The TwitchExtension allows to display the actual PowerTips of Memories and the Snap Counter on stream

- The data is shared through two static ArrayLists: slayTheRelicsHitboxes, slayTheRelicsPowerTips

- The specific properties are looked up (by name) by the SlayTheRelicsExporter mod (in all classes that have @SpireInitializer), if they are found, they are checked frequently and their contents are sent over to a server that communicates with Twitch

- Here, during every update of the Wanderer both ArrayLists are cleared and refilled. This is perhaps not the most optimal approach but probably the simplest with the smallest footprint. Impact on performance should be negligable.

- Documentation for the API will be available soon (tm), the extension is expected to be released within 1-3 weeks (2020/4/5 - 2020/4/19)

- The changes **have been tested** reasonably with mentioned mod and extension directly on Twitch.

### DIFF
--- a/src/main/java/stsjorbsmod/characters/Wanderer.java
+++ b/src/main/java/stsjorbsmod/characters/Wanderer.java
@@ -474,10 +474,23 @@ public class Wanderer extends CustomPlayer implements OnResetPlayerSubscriber {
 
     @Override
     public void update() {
-        TwitchExtensionAPI.clearLists();
         super.update();
         int flipMultiplier = flipHorizontal ? -1 : 1;
         snapCounter.update(drawX + flipMultiplier * SNAP_COUNTER_OFFSET_X, drawY + SNAP_COUNTER_OFFSET_Y, flipMultiplier);
+        updateTwitchExtensionAPI();
+    }
+
+    private void updateTwitchExtensionAPI() {
+        TwitchExtensionAPI.clearLists();
+
+        MemoryManager manager = MemoryManager.forPlayer();
+        if (!manager.isSnapped()) {
+            for (int i = 0; i < manager.memories.size(); ++i) {
+                TwitchExtensionAPI.addPowerTips(manager.memories.get(i).hb, manager.memories.get(i).getTips());
+            }
+
+            TwitchExtensionAPI.addPowerTips(snapCounter.hb, snapCounter.tips);
+        }
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


avolny:
Please hang on. This works fine when one has loaded only the Wanderer mod, problem arised when I loaded Poker Player as well and tried to play it. The problem is that MemoryManager constructor and update calls are patched directly into AbstractPlayer. This means that when the PokerPlayer instance is created, it contains MemoryManager and in each update step it adds its memories to the TwitchExtensionAPI. The shared list gets emptied at the beginning of each update in the Wanderer class but it doesn't get emptied every update in the PokerPlayer class, that's why it grew indefinitely. But in other characters this code should not even be called at all. This would probably be the same case even with the base game characters, they will have the MemoryManager patched in as well, I believe.

Do you have suggestions on how to fix this behavior easily?

My idea is to make all the used properties public (MemoryManager.memories, SnapCounter.tips, SnapCounter.hitbox) and do the whole TwitchExtensionAPI update in a single place - in the Wanderer update() function instead of it being delegated to multiple places.

dbjorge:
I would recommend doing it in the other direction and avoiding having `Wanderer` be responsible for any aspect of managing the memory hitbox/tip lists; `MemoryManager` is patched into `AbstractPlayer` rather than being a member of `Wanderer` intentionally because it is possible for non-Wanderer classes to be rendering memories (in particular, CULL always renders one memory, and any class which somehow gains access to a Wanderer card via Prismatic-Shard-like shenanigans will start rendering those memories which are currently active).

I do think it makes sense for the individual things with hitboxes/tips to be responsible for making the decisions about whether they should be rendered or not, same as they're responsible for determining whether they should be rendered on screen or not. Personally, I would probably lean towards something like:

```java
@SpireInitializer
public class SlayTheRelicsIntegration {
    public static ArrayList<Hitbox> slayTheRelicsHitboxes = new ArrayList<>();
    public static ArrayList<ArrayList<PowerTip>> slayTheRelicsPowerTips = new ArrayList<>();

    public static void initialize() {
        BaseMod.subscribe((PreRenderSubscriber) () -> SlayTheRelicsIntegration.clear());
    }

    public static void clear() {
        slayTheRelicsHitboxes.clear();
        slayTheRelicsPowerTips.clear();
    }

    public static void renderTipHitbox(Hitbox hb, ArrayList<PowerTip> tips) {
        slayTheRelicsHitboxes.add(hb);
        slayTheRelicsPowerTips.add(tips);
    }
}
```

...and then have `AbstractMemory::renderTip` and `SnapCounter::render` be responsible for calling `SlayTheRelicsIntegration.renderTipHitbox` appropriately.

avolny:
What you suggest seems like a good approach. It lead to a lot less code than in my original implementation and is more general - works even on other characters using jorbs-mod cards.

New iteration of changes discussed in #485 

Closing this PR.

PR #502, opened 2020-05-01
Ticking curse

acourrau:
Adds Ticking Curse card for the Deck of Trials. Special calculation is done for the second attack damage value. It's held by magicNumber, but applied as damage.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -749,6 +749,10 @@
     "DESCRIPTION": "stsjorbsmod:Ephemeral. NL Advance ALL buffs and debuffs !M! turn through time.",
     "UPGRADE_DESCRIPTION": "stsjorbsmod:Ephemeral. NL Advance ALL buffs and debuffs !M! turns through time."
   },
+  "stsjorbsmod:TickingCurse": {
+    "NAME": "Ticking Curse",
+    "DESCRIPTION": "Ethereal. NL Deal !D! damage to all enemies. stsjorbsmod:Exert. NL Deal !M! damage to all enemies when this exhausts."
```
dbjorge, line 754: I think usually conditional effects like the last line try to state the condition first; maybe `When exhausted, deal !M! damage to ALL enemies.`?

dbjorge, line 754: The base game makes the "ALL" all-caps whenever it uses the phrase `to ALL enemies`; would be good to be consistent

src/main/java/stsjorbsmod/cards/cull/deckoftrials/TickingCurse.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        this.addToBot(new DamageAllEnemiesAction(p, DAMAGE, this.damageType, AbstractGameAction.AttackEffect.SLASH_HORIZONTAL));
+    }
+
+    public void triggerOnExhaust() {
+        if (this.upgraded) {
+            this.addToBot(new DamageAllEnemiesAction(null, EXHAUST_DAMAGE + UPGRADE_EXHAUST_DAMAGE, this.damageType, AbstractGameAction.AttackEffect.FIRE));
```
dbjorge, line 41: I think this is ignoring the (correct-looking) calculation in the custom calculateCardDamage/applyPowers impls; I think this should be using `magicNumber` and doesn't have to do a `this.upgraded` check

src/main/java/stsjorbsmod/cards/cull/deckoftrials/TickingCurse.java
```diff
@@ -0,0 +1,95 @@
...
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        baseDamage = DAMAGE;
+        magicNumber = baseMagicNumber = EXHAUST_DAMAGE;
+        this.isEthereal = true;
+        SelfExertField.selfExert.set(this, true);
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        this.addToBot(new DamageAllEnemiesAction(p, DAMAGE, this.damageType, AbstractGameAction.AttackEffect.SLASH_HORIZONTAL));
```
dbjorge, line 36: This should be using `this.multiDamage` rather than `DAMAGE`; multiDamage is set by `calculateCardDamage` to handle cases where different enemies have different damage numbers due to different vulnerable/etc statuses

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
acourrau: Agree.

acourrau: Agree.

src/main/java/stsjorbsmod/cards/cull/deckoftrials/TickingCurse.java
acourrau: Agree

src/main/java/stsjorbsmod/cards/cull/deckoftrials/TickingCurse.java
acourrau: Agree.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -749,6 +749,10 @@
     "DESCRIPTION": "stsjorbsmod:Ephemeral. NL Advance ALL buffs and debuffs !M! turn through time.",
     "UPGRADE_DESCRIPTION": "stsjorbsmod:Ephemeral. NL Advance ALL buffs and debuffs !M! turns through time."
   },
+  "stsjorbsmod:TickingCurse": {
+    "NAME": "Ticking Curse",
+    "DESCRIPTION": "Ethereal. NL Deal !D! damage to all enemies. stsjorbsmod:Exert. NL When exhausted, deal !M! damage to ALL enemies."
```
dbjorge, line 754: "Exhausted" should be capitalized for consistency with the base game (eg, Dark Embrace/Fiend Fire/Feel No Pain)

src/main/java/stsjorbsmod/cards/cull/deckoftrials/TickingCurse.java
```diff
@@ -0,0 +1,53 @@
...
+        this.isEthereal = true;
+        SelfExertField.selfExert.set(this, true);
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        this.addToBot(new DamageAllEnemiesAction(p, this.multiDamage, this.damageType, AbstractGameAction.AttackEffect.SLASH_HORIZONTAL));
+    }
+
+    public void triggerOnExhaust() {
```
dbjorge, line 40: It's a good practice to put `@Override` anytime you're overriding a base class method; it doesn't break anything to omit it, but it makes it easier for other programmers to understand your intent and it helps prevent typo'ing a method name or a parameter type in a way that would be confusing to debug.

acourrau: Whoops, missed that. Fixed.

PR #503, opened 2020-05-01
Prodigal memory

acourrau:
Added Prodigal Memory card to Deck of Trials. Special logic for Exerting the card if not played. Perhaps we should consider adding an ExertAction if we want to explicitly call Exert in other places. For now it appears one-off so I think it's okay as is.


src/main/java/stsjorbsmod/cards/cull/deckoftrials/ProdigalMemory.java
```diff
@@ -0,0 +1,60 @@
...
+public class ProdigalMemory extends CustomJorbsModCard {
+    public static final String ID = JorbsMod.makeID(ProdigalMemory.class);
+
+    private static final CardRarity RARITY = CardRarity.SPECIAL;
+    private static final CardTarget TARGET = CardTarget.SELF;
+    private static final CardType TYPE = CardType.SKILL;
+    public static final CardColor COLOR = Cull.Enums.CULL_CARD_COLOR;
+
+    private static final int COST = 0;
+    private static final int DRAW = 3; // Hard coded in localization strings too.
```
dbjorge, line 25: Best to avoid hard-coding it in the localization strings; you can use `urMagicNumber` or `metaMagicNumber` to avoid that

src/main/java/stsjorbsmod/cards/cull/deckoftrials/ProdigalMemory.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster abstractMonster) {
+        if (this.dontTriggerOnUseCard) {
+            SelfExertField.selfExert.set(this, true);
+            addToBot(new ExhaustSpecificCardAction(this, AbstractDungeon.player.hand));
+        }
+        else {
+            addToBot(new DrawCardAction(DRAW));
+            AbstractDungeon.actionManager.addToBottom(
```
dbjorge, line 42: This can just be `addToBot`, it's shorthand for `AbstractDungeon.actionManager.addToBottom`

src/main/java/stsjorbsmod/cards/cull/deckoftrials/ProdigalMemory.java
```diff
@@ -0,0 +1,60 @@
...
+                    new DecreaseMaxHpAction(p, p, magicNumber, AbstractGameAction.AttackEffect.POISON));
+        }
+    }
+
+    public void triggerOnEndOfTurnForPlayingCard() {
```
dbjorge, line 47: Missing `@Override`

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -586,6 +586,10 @@
     "NAME": "Procrastination",
     "DESCRIPTION": "Unplayable. NL Starts each combat in your discard pile."
   },
+  "stsjorbsmod:ProdigalMemory": {
+    "NAME": "Prodigal Memory",
+    "DESCRIPTION": "Draw 3 cards and lose !M! max HP. NL stsjorbsmod:Exert if not played."
```
dbjorge, line 591: Wording on the "exert if not played" should probably be a little closer to how Burn/Shame/Regret/etc are worded. They all use `At the end of your turn, <do a thing>.` to make it clear when exactly the effect happens, but they're also unplayable, so I do think it's good for us to include some sort of conditional wording too. Maybe something like `If this card is in your hand at the end of your turn, stsjorbsmod:Exert.`?

PR #504, opened 2020-05-02
Investment

acourrau:
Added the Investment card to Deck of Trials. Has logic to count how many times it's been drawn. Used the same Description logic as FTL to display how many times it's been drawn dynamically.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -412,6 +412,15 @@
     "NAME": "Introspection",
     "DESCRIPTION": "At the end of each turn, lose !stsjorbsmod:UrMagic! HP and deal !M! damage to ALL enemies. Deal !stsjorbsmod:MetaMagic! additional damage per stsjorbsmod:Clarity."
   },
+  "stsjorbsmod:Investment": {
+    "NAME": "Investment",
+    "DESCRIPTION": "stsjorbsmod:Ephemeral. Gain !M! gold for each time this has been drawn. NL stsjorbsmod:Exert.",
```
dbjorge, line 417: We usually use an `NL` to separate "Ephemeral" to its own line, similar to how the base game usually does for "Ethereal"

dbjorge: I think the text might be more clear if we include the term "permanently" somehow, like the base game cards with effects spanning multiple combats (Ritual Dagger, Genetic Algorithm" usually do). Maybe something like:

`stsjorbsmod:Ephemeral. NL Gain !M! gold. stsjorbsmod:Exert. NL When you draw this card, permanently increase its gold gain by !stsjorbsmod:metaMagic!.`

Note, this is intentionally embedding the current amount in the main text (like Genetic Algorithm and Ritual Dagger do) rather than using an extended description; I think for cards with permanent effects that's probably a better pattern, since you want the current number to still be visible in the master deck view outside of combat.

dbjorge: (if you do this, you'll still want to use `misc` for the sake of making the value saveable, but you'll need to start overriding `applyLoadedMiscValue` and `applyPowers` to sync a magicNumber to the misc value

src/main/java/stsjorbsmod/cards/cull/deckoftrials/Investment.java
```diff
@@ -0,0 +1,64 @@
...
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        baseMagicNumber = magicNumber = GOLD_PER_DRAW;
+
+        this.misc = 0;
+
+        EphemeralField.ephemeral.set(this, true);
+        SelfExertField.selfExert.set(this, true);
+    }
+
+    public void triggerWhenDrawn() {
```
dbjorge, line 37: `@Override`

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
acourrau: Agreed, makes the card feel cleaner/clearer.

PR #505, opened 2020-05-02
Blasphemer

nl364:


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -59,6 +59,10 @@
     "NAME": "Black Tentacles",
     "DESCRIPTION": "Deal !D! damage. NL This turn, all damage you deal to other enemies is dealt to this enemy instead."
   },
+  "stsjorbsmod:Blasphemer": {
+    "NAME": "Blasphemer",
+    "DESCRIPTION": "Upon draw, add !M! *Smites into your hand"
```
dbjorge, line 64: I think this should probably mention "Unplayable" and the self-exhaust effect. Maybe something like:

`Unplayable. NL When drawn, add !M! *Smites into your hand and Exhaust.`

src/main/java/stsjorbsmod/cards/cull/deckoftrials/Blasphemer.java
```diff
@@ -0,0 +1,53 @@
...
+import stsjorbsmod.JorbsMod;
+import stsjorbsmod.cards.CustomJorbsModCard;
+import stsjorbsmod.characters.Cull;
+
+public class Blasphemer extends CustomJorbsModCard {
+    public static final String ID = JorbsMod.makeID(Blasphemer.class);
+
+    private static final CardRarity RARITY = CardRarity.SPECIAL;
+    private static final CardTarget TARGET = CardTarget.SELF;
+    private static final CardType TYPE = CardType.SKILL;
```
dbjorge, line 18: I think SKILL is correct based on the design sheet, but it looks like you may have run the generate-card-images script when this was marked as a power; can you re-run it?

src/main/java/stsjorbsmod/cards/cull/deckoftrials/Blasphemer.java
```diff
@@ -0,0 +1,53 @@
...
+    public static final CardColor COLOR = Cull.Enums.CULL_CARD_COLOR;
+
+    private static final int COST = -2;
```
dbjorge, line 21: We added a `COST_UNPLAYABLE` constant to make this a little more readable, see `Amnesia.java` for a usage example

src/main/java/stsjorbsmod/cards/cull/deckoftrials/Blasphemer.java
```diff
@@ -0,0 +1,53 @@
...
+    public Blasphemer() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        this.exhaust = true;
+        this.cardsToPreview = new Smite();
+        magicNumber = baseMagicNumber = SMITE_AMOUNT;
+    }
+
+    @Override
+    public void triggerWhenDrawn() {
+        this.addToTop(new MakeTempCardInHandAction(new Smite(), this.magicNumber));
```
dbjorge, line 34: Usually for a card, you want to use `addToBot` unless there's some specific exceptional reason to use `addToTop`. Actions that invoke other actions are usually the only place where `addToTop` is the more appropriate default choice.

src/main/java/stsjorbsmod/cards/cull/deckoftrials/Blasphemer.java
```diff
@@ -0,0 +1,53 @@
...
+            this.upgradeName();
+            this.upgradeMagicNumber(UPGRADE_SMITE_AMOUNT);
+        }
+    }
+
+    @Override
+    public void use(AbstractPlayer abstractPlayer, AbstractMonster abstractMonster) {
+    }
+
+    public boolean canUse(AbstractPlayer p, AbstractMonster m) {
```
dbjorge, line 50: It's a good practice to put `@Override` anytime you're overriding a base class method; it doesn't break anything to omit it, but it makes it easier for other programmers to understand your intent and it helps prevent typo'ing a method name or a parameter type in a way that would be confusing to debug.

src/main/java/stsjorbsmod/cards/cull/deckoftrials/Blasphemer.java
```diff
@@ -0,0 +1,53 @@
...
+    public void triggerWhenDrawn() {
+        this.addToTop(new MakeTempCardInHandAction(new Smite(), this.magicNumber));
+        this.addToTop(new ExhaustSpecificCardAction(this, AbstractDungeon.player.hand));
+    }
+
+    @Override
+    public void upgrade() {
+        if (!this.upgraded) {
+            this.upgradeName();
+            this.upgradeMagicNumber(UPGRADE_SMITE_AMOUNT);
```
dbjorge, line 42: We like to include `this.upgradeDescription()` in all cards' `upgrade` methods; it doesn't really cost anything noticeable to do it when it's not needed, and it's a good practice because there are some rare cases where English doesn't need a special upgrade description but a different language might want one.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
dbjorge: Asked jorbs to confirm the "unplayable" in discord (I agree with you implementing it that way based on my interpretation of the sheet, but want to double check)

src/main/java/stsjorbsmod/cards/cull/deckoftrials/Blasphemer.java
```diff
@@ -0,0 +1,53 @@
+package stsjorbsmod.cards.cull.deckoftrials;
```
dbjorge, line 1: Because this is a Deck of Trials card, we'll need to update `DeckOfTrials.java`'s `reset()` to add it to the deck

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
dbjorge: Jorbs pointed out that Deus Ex Machina is a very similar example and we should follow it. I think the behavior already follows it, but that example suggests that the text should probably be:

`"Unplayable. NL When you draw this card, add !M! *Smites to your hand and Exhaust."`

src/main/java/stsjorbsmod/cards/cull/deckoftrials/Blasphemer.java
nl364: I thought I did put in in the Deck of Trials reset(). Well, regardless, it should be there in the next pr

PR #506, opened 2020-05-02
Shrapnel bloom

acourrau:
Added Shrapnel Bloom card to the Deck of Trials. Copied and tweaked logic for duplication power. Exert played card is again one-off, should probably have it's own action, but I abandoned that for the time being. Works correctly for now.


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -241,6 +241,14 @@
       "Reduce healing by %1$s%%."
     ]
   },
+  "stsjorbsmod:ShrapnelBloomPower": {
+    "NAME": "Shrapnel Bloom",
+    "DESCRIPTIONS": [
+      "This turn, your next",
```
dbjorge, line 247: I don't think mixing the stacking behavior does quite the right thing; if you somehow have a Shrapnel Bloom and a Shrapnel Bloom+ and play both, I think what's currently going to happen is the number of card will stack but the number of plays per card will use the first value and ignore the second. Asked jorbs to confirm the desired stacking behavior in discord.

dbjorge: Per discord, let's go with the same thing the Wanderer's `Gather Power` does, where "stacking increases the number of times the next one card is played, it never affects multiple next cards". That probably suggests we should update the wording on the card/power to be more consistent with Gather Power and say `This turn, your next non-Legendary card is played !M! additional time.` rather than `...played 2 times`, since it makes the stacking behavior more obvious to a player.

src/main/java/stsjorbsmod/powers/ShrapnelBloomPower.java
```diff
@@ -0,0 +1,93 @@
...
+        } else {
+            this.description += String.format(DESCRIPTIONS[2], this.amount, numPlays);
+        }
+    }
+
+    @Override
+    public AbstractPower makeCopy() { return new ShrapnelBloomPower(owner, amount, numPlays); }
+
+    @Override
+    public boolean onDamageToRedirect(AbstractPlayer player, DamageInfo info, AbstractGameAction.AttackEffect effect) {
```
dbjorge, line 90: You can omit this and the `implements OnDamageToRedirectSubscriber` at the top :)

src/main/java/stsjorbsmod/powers/ShrapnelBloomPower.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public void onUseCard(AbstractCard card, UseCardAction action) {
+        if (!card.purgeOnUse && this.amount > 0 && !card.hasTag(LEGENDARY)) {
+            this.flash();
+
+            action.exhaustCard = true; // this is what corruption does
+            SelfExertField.selfExert.set(card, true);
+
+            AbstractMonster m = null;
```
dbjorge, line 42: You can simplify L42-L60 a bit similar to what we've already done for `PlayNextAttackThisTurnAdditionalTimesPower`:

```java
            AbstractMonster m = (AbstractMonster)action.target;

            for (int i = 0; i < amount; ++i) {
                CardMetaUtils.playCardAdditionalTime(card, m);
            }
```

src/main/java/stsjorbsmod/powers/ShrapnelBloomPower.java
```diff
@@ -0,0 +1,93 @@
...
+        super(STATIC);
+
+        this.owner = owner;
+        this.amount = amount;
+        this.numPlays = numPlays;
+
+        updateDescription();
+    }
+
+    public void onUseCard(AbstractCard card, UseCardAction action) {
```
dbjorge, line 35: A few `@Overrides`s

src/main/java/stsjorbsmod/powers/ShrapnelBloomPower.java
```diff
@@ -0,0 +1,77 @@
...
+            action.exhaustCard = true; // this is what corruption does
+            SelfExertField.selfExert.set(card, true);
+
+            AbstractMonster m = (AbstractMonster)action.target;
+
+            for (int i = 0; i < amount; ++i) {
+                CardMetaUtils.playCardAdditionalTime(card, m);
+            }
+
+            --this.amount;
```
dbjorge, line 48: I think this is still based on amount being the "number of next cards", rather than "number of additional plays"; I think now that it's the latter we probably want this to just unilaterally remove itsself after use

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -675,6 +675,11 @@
     "NAME": "Shake",
     "DESCRIPTION": "If enemy is Vulnerable, draw !M! cards. NL Apply !stsjorbsmod:UrMagic! Vulnerable. NL Exhaust."
   },
+  "stsjorbsmod:ShrapnelBloom": {
+    "NAME": "Shrapnel Bloom",
+    "DESCRIPTION": "stsjorbsmod:Exert. NL This turn, the next non-Legendary card you play is played !M! additional time, then exerts.",
```
dbjorge, line 680: The term "exerts" at the end of the 2 descriptions should probably also be capitalized/colorized like a keyword (`stsjorbsmod:Exerts` here, `#yExerts` in the power description)

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -241,6 +241,13 @@
       "Reduce healing by %1$s%%."
     ]
   },
+  "stsjorbsmod:ShrapnelBloomPower": {
+    "NAME": "Shrapnel Bloom",
+    "DESCRIPTIONS": [
+      "This turn, your next non-legendary card is played %1d additional time, then #yexerts.",
```
dbjorge, line 247: I don't think we can colorize just the "legendary" part of "non-legendary", but probably best to at least capitalize it like a keyword (and for consistency with the card text does)

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
acourrau: "Exerts" was not a Keyword, added that and updated.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
acourrau: Fixed.

src/main/java/stsjorbsmod/powers/ShrapnelBloomPower.java
acourrau: Yup missed that, fixed.

PR #508, opened 2020-05-04
Assertion

acourrau:
Added Assertion card to Deck of Trials. Some special logic to cap base damage at 50(75). New save data (skills played per act). JorbsMod change to check every card played if it's a skill to increment counter appropriately.


src/main/java/stsjorbsmod/JorbsMod.java
```diff
@@ -440,6 +439,9 @@ public void receiveStartAct() {
             }
             ExertedField.exerted.set(c, false);
         }
+        SkillsPlayedThisActSaveData sp =
+                (SkillsPlayedThisActSaveData)BaseMod.getSaveFields().get(MOD_ID + ":SkillsPlayedThisAct");
+        sp.skillsPlayed = 0;
```
dbjorge, line 444: This is just immediately discarding whatever `BaseMod.getSaveFields().get` is loading, no? Can this not juts be `SkillsPlayedThisActSaveData.skillsPlayed = 0;`?

src/main/java/stsjorbsmod/JorbsMod.java
```diff
@@ -456,4 +458,14 @@ public void receiveStartGame() {
             }
         }
     }
+
+    @Override
+    public void receiveCardUsed(AbstractCard c) {
+        SkillsPlayedThisActSaveData sp =
```
dbjorge, line 464: Same here; I don't think `BaseMod.getSaveFields` is necessary, the `skillsPlayed` field is static

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -51,6 +51,16 @@
     "NAME": "Arcane Weapon",
     "DESCRIPTION": "At the end of each turn, deal !D! stsjorbsmod:Attack\u00a0Damage to a random enemy."
   },
+  "stsjorbsmod:Assertion": {
+    "NAME": "Assertion",
+    "DESCRIPTION": "Deal ",
```
dbjorge, line 56: It's preferable to use magicNumbers rather than string concatenation for filling in dynamic values; it's more localization-friendly. Combined with the feedback above about using `getRawDynamicDescriptionSuffix()`, I'd recommend something like:

```jsonc
    "DESCRIPTION": "Deal damage equal to !M! times the number of skills played this act, up to !stsjorbsmod:metaMagic! damage.",
    "EXTENDED_DESCRIPTION": [
      " NL (Deals !D! damage, !stsjorbsmod:urMagic! skills played.)"
    ]
```

src/main/java/stsjorbsmod/cards/cull/deckoftrials/Assertion.java
```diff
@@ -0,0 +1,82 @@
...
+        addToBot(new DamageAction(m, new DamageInfo(p, damage), AbstractGameAction.AttackEffect.SLASH_HORIZONTAL));
+    }
+
+    @Override
+    public void applyPowers() {
+        super.applyPowers();
+        SkillsPlayedThisActSaveData sp =
+                (SkillsPlayedThisActSaveData)BaseMod.getSaveFields().get(MOD_ID + ":SkillsPlayedThisAct");
+
+        this.rawDescription = cardStrings.DESCRIPTION;
```
dbjorge, line 54: We've had to do these rawDescription shenanigans often enough that we abstracted it somewhat in `CustomJorbsModCard`. You can omit the custom `applyPowers` and instead do something like this (combined with the suggested text change):

```java
    @Override
    public int calculateBonusMagicNumber() {
        return SkillsPlayedThisActSaveData.skillsPlayed;
    }
    
    @Override
    public String getRawDynamicDescriptionSuffix() {
        return EXTENDED_DESCRIPTION[0];
    }
```

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
dbjorge: To do the skill/skills pluralization, you could still use 2 extended description options, but it's more localization friendly to accept a little repetition to avoid the concatenation, so something like this:

```jsonc
    "EXTENDED_DESCRIPTION": [
      " NL (Deals !D! damage, !M! skills played.)",
      " NL (Deals !D! damage, !M! skills played.)"
    ]
```

...and have `getRawDynamicDescriptionSuffix return one or the other based on `magicNumber` being == 1

PR #509, opened 2020-05-05
Repressed Memory - small bug where looking at discard pile reduces max hp

nic3245:


src/main/java/stsjorbsmod/actions/RepressedMemoryAction.java
```diff
@@ -0,0 +1,35 @@
...
+    public void update() {
+        //if (duration == baseDuration) {
+          //  AbstractDungeon.effectList.add(new FlashAtkImgEffect(this.target.hb.cX, this.target.hb.cY, AttackEffect.BLUNT_HEAVY));
+       // }
+        //tickDuration();
+        //if (isDone) {
+            if (CombatUtils.isCombatBasicallyVictory()) {
+                addToBot(new ExertAction(this.card));
+                JorbsMod.logger.info("Did ExertAction. Exerted? " + ExertedField.exerted.get(this.card));
+                isDone = true;
```
wang429, line 31: If the combat is not "basically victory" this line will never trigger and thus cause the ActionManager to infinitely evaluate this action.

wang429, line 31: This line should be moved out and after the conditional. Otherwise, the action manager will infinitely call `update()`

src/main/java/stsjorbsmod/actions/RepressedMemoryAction.java
```diff
@@ -0,0 +1,35 @@
...
+import com.megacrit.cardcrawl.actions.AbstractGameAction;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.core.Settings;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.vfx.combat.FlashAtkImgEffect;
+import stsjorbsmod.JorbsMod;
+import stsjorbsmod.patches.ExertedField;
+import stsjorbsmod.util.CombatUtils;
+
+public class RepressedMemoryAction extends AbstractGameAction {
```
wang429, line 12: I'm not sure this action is altogether necessary. Either just do everything in the card's `use` or in the card's action. I prefer the former.

src/main/java/stsjorbsmod/cards/cull/deckoftrials/RepressedMemory.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Override
+    public void use(AbstractPlayer abstractPlayer, AbstractMonster abstractMonster) {
+        addToBot(new DrawCardAction(this.magicNumber));
+        addToBot(new GainEnergyAction(this.urMagicNumber));
+        addToBot(new RepressedMemoryAction(this));
+        //line below would use power instead
+        addToBot(new ApplyPowerAction(abstractPlayer, abstractPlayer, new RepressedMemoryPower(abstractPlayer, this)));
+        JorbsMod.logger.info("Exerted? " + ExertedField.exerted.get(this));
+        //JorbsMod.logger.info("SelfExerted? " + SelfExertField.selfExert.get(this));
```
wang429, line 61: I think using SelfExertedField is more consistent than relying on ExertAction. A potential issue of ExertAction might be that it should always get cleared if the combat ends while the UseCardAction is active.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -629,6 +629,10 @@
     "NAME": "Refuse to Forget",
     "DESCRIPTION": "You cannot forget your current memory until you have stsjorbsmod:Clarity of it. NL Draw !M! cards. NL You cannot draw more cards this turn."
   },
+  "stsjorbsmod:RepressedMemory": {
+    "NAME": "Repressed Memory",
+    "DESCRIPTION": "Lose !stsjorbsmod:MetaMagic! max hp when discarded NL If played: Draw !M! cards. Gain !stsjorbsmod:UrMagic! [E]. Exert at end of fight."
```
wang429, line 634: is `stsjorbsmod:Exert` a keyword? If so, please use it.

dbjorge:
Per nl364 in discord, deferring review for the moment:

> I’m away from computer, but I just realized RepressedMemory needs a separate thing on start of act. Kinda just making a note/saying don’t look at the pr again yet

acourrau:
The debuff should only happen when discarded from hand (end of turn or manually). See response from jorbs:
https://discordapp.com/channels/329758756076519436/638465421598654496/770289567063998464

I think all we need to do is override the `triggerOnManualDiscard()` and `triggerOnEndOfPlayerTurn()`. For the later, we need to check if player has runic pyramid. The rest should be fine I think.

acourrau:
Also I think you have the same issue here with branch names as with posession, one is `repressedMemory` one is `repressedmemory`, the later looks like it is more up to date.

acourrau:
Closing PR so that I can submit one from my repo with master merges/fixes/etc

PR #510, opened 2020-05-06
Mirrored technique

acourrau:
Added Mirrored Technique, new persistent power for combat that will count the number of incoming attacks, and then play the card that many times. Perhaps this should apply the damage that many times because it currently procs relics like kunai, ink bottle, nunchaku, etc. 

Also the Reflection to get intentMultiAmt could probably use the ReflectionUtils method, but I'm unsure what it would look like.

I have a feeling this card will get balanced. It's very very strong.


src/main/java/stsjorbsmod/powers/MirroredTechniquePower.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private static Field multiIntentField;
+    private static int extraPlays;
+
+    public MirroredTechniquePower(final AbstractCreature owner, final int extraPlays) {
+        super(STATIC);
+
+        this.owner = owner;
+        this.isTurnBased = false;
+        this.extraPlays = extraPlays;
```
dbjorge, line 27: I think we'll want to consider the desired stacking behavior here; depending on design feedback, we might want to use amount for this if it's intended to stack on multiple Mirrored Technique+ usages, or if it's not, this will need custom stacking behavior to take the Math.max value on stack (to cover both orders of Mirrored Technique and Mirrored Technique+)

src/main/java/stsjorbsmod/powers/MirroredTechniquePower.java
```diff
@@ -0,0 +1,91 @@
...
+
+        updateDescription();
+    }
+
+    @Override
+    public AbstractPower makeCopy() {
+        return new MirroredTechniquePower(owner, amount);
```
dbjorge, line 34: This is using `amount`, not `extraPlays`

src/main/java/stsjorbsmod/powers/MirroredTechniquePower.java
```diff
@@ -0,0 +1,91 @@
...
+            for (int i = 0; i < multiUse + extraPlays; ++i) {
+                CardMetaUtils.playCardAdditionalTime(card, m);
+            }
+        }
+    }
+
+    public static int getIncomingAttackCount() {
+        int totalCount = 0;
+
+        // Some way to turn this into a call to ReflectionUtils.getPrivateField(...)
```
dbjorge, line 53: `ReflectionUtils.getPrivateField(monster, AbstractMonster.class, "intentMultiAmt")` is the syntax.

src/main/java/stsjorbsmod/powers/MirroredTechniquePower.java
```diff
@@ -0,0 +1,91 @@
...
+                    }
+                    totalCount += multiAmt;
+                }
+            }
+        }
+
+        return totalCount;
+    }
+
+    public static boolean isAttacking(AbstractMonster m) {
```
dbjorge, line 78: We have a helper for this already at `IntentUtils.isAttackIntent`

src/main/java/stsjorbsmod/powers/MirroredTechniquePower.java
```diff
@@ -0,0 +1,91 @@
...
+        if(AbstractDungeon.getMonsters() != null) {
+            for (AbstractMonster m : AbstractDungeon.getMonsters().monsters) {
+                if (!m.isDeadOrEscaped() && isAttacking(m)) {
+                    int multiAmt = 0;
+                    try {
+                        multiAmt = (int) multiIntentField.get(m);
+                    }
+                    catch (IllegalAccessException ignored) {
+                    }
+                    totalCount += multiAmt;
```
dbjorge, line 70: I don't think this does the right thing in the event that the enemy *isn't* multi-attacking; should probably wrap in a check that uses `1` if `m.isMultiDmg` is false.

acourrau:
Updated with PR comments and added Exert functionality.

src/main/java/stsjorbsmod/powers/MirroredTechniquePower.java
```diff
@@ -0,0 +1,79 @@
...
+import com.megacrit.cardcrawl.powers.AbstractPower;
+import stsjorbsmod.util.CardMetaUtils;
+import stsjorbsmod.util.IntentUtils;
+import stsjorbsmod.util.ReflectionUtils;
+
+public class MirroredTechniquePower extends CustomJorbsModPower {
+    public static final StaticPowerInfo STATIC = StaticPowerInfo.Load(MirroredTechniquePower.class);
+    public static final String POWER_ID = STATIC.ID;
+
+    private static int extraPlays;
```
dbjorge, line 17: I think it would be preferable for this to be non-static; this feels like it should be part of a given instance of the power to me, not shared among all instances. To achieve the desired stacking behavior, you'd want to pair "making this nonstatic" with moving that `Math.max` call out of the ctor and into an override of the `stackPower` method

acourrau: Makes sense, thanks.

PR #511, opened 2020-05-07
Possession

nic3245:


src/main/java/stsjorbsmod/cards/cull/deckoftrials/Possession.java
```diff
@@ -0,0 +1,61 @@
...
+        int count = 0;
+        if (AbstractDungeon.getMonsters() != null) {
+            for (AbstractMonster monster : AbstractDungeon.getMonsters().monsters) {
+                if (!monster.isDead && !monster.isDying) {
+                    count += 1;
+                }
+            }
+        }
+        if (count != 1) {
+            addToBot(new ApplyPowerAction(m, p, new PossessionPower(m, this.magicNumber)));
```
dbjorge, line 44: I think for the purposes of interacting with artifact, it would probably make more sense for the "possession" effect and the "stun" effect to be treated as one combined debuff rather than eating two different charges of artifact. To achieve that, I'd recommend using a pattern similar to how `BanishedPower` works:

* Always have the card apply `PossessionPower`, regardless of number of enemies
* Move the application of `StunMonsterPower` from here into an `onInitialApplication` hook in `PossessionPower`
* Move the "ignore the damage if there aren't any monsters" logic to be part of the target selection logic in `PossessionPower:atStartOfTurn`

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -567,6 +567,10 @@
       " NL (You have !M! Curses.)"
     ]
   },
+  "stsjorbsmod:Possession": {
+    "NAME": "Possession",
+    "DESCRIPTION": "Possess target, dealing it's max hp in damage !M! times to another random target at the end of your turn. NL stsjorbsmod:Exert"
```
dbjorge, line 572: `it's` should be `its`

dbjorge, line 572: The game usually capitalizes the term "HP"

dbjorge, line 572: Descriptions should end in periods

dbjorge, line 572: I think maybe the wording should be more clear that it's a debuff on the enemy and that the enemy will be attacking its ally *instead* of performing its normal action next turn. Maybe something like:

```Apply !M! stsjorbsmod:Possessed. NL stsjorbsmod:Exert.```

...with a new keyword for possessed along the lines of:

```#yPossessed creatures will spend their next turn attacking random allies. They will attack a number of times equal to their #yPossessed stacks. Each attack deals damage equal to their max HP.```

File: src/main/java/stsjorbsmod/powers/PossessionPower.java [hunk 3]
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public void updateDescription() {
+        description = this.amount == 1 ? DESCRIPTIONS[0] : String.format(DESCRIPTIONS[1], this.amount);
+    }
+
+    @Override
+    public AbstractPower makeCopy() {
+        return null;
+                //new PossessionPower(owner, amount);
```
dbjorge, line 60: I think the commented version is correct here; `makeCopy` is important for being able to apply the power via the dev console for testing

src/main/java/stsjorbsmod/powers/PossessionPower.java
```diff
@@ -0,0 +1,63 @@
...
+        updateDescription();
+    }
+
+    @Override
+    public void atStartOfTurn() {
+        AbstractMonster m = this.target;
+        JorbsMod.logger.info("m and this.owner: " + m);
+
+        while (m == this.target) {
+            m = AbstractDungeon.getMonsters().getRandomMonster((AbstractMonster) null, true, AbstractDungeon.cardRandomRng);
```
dbjorge, line 34: We have a helper in `CombatUtils` that I think would make this easier to implement and less prone to infinite loops:

```java
m = CombatUtils.getRandomAliveMonster(AbstractDungeon.getMonsters(), candidate -> candidate != this.owner, AbstractDungeon.cardRandomRng);
```

src/main/java/stsjorbsmod/powers/PossessionPower.java
```diff
@@ -0,0 +1,63 @@
...
+    public static final StaticPowerInfo STATIC = StaticPowerInfo.Load(PossessionPower.class);
+    public static final String POWER_ID = STATIC.ID;
+    private final AbstractMonster target;
+
+    public PossessionPower(AbstractMonster owner, int amount) {
+        super(STATIC);
+        this.type = PowerType.DEBUFF;
+        this.owner = owner;
+        this.amount = amount;
+        this.target = owner;
```
dbjorge, line 22: I don't think the separate `target` field is necessary here; it feels misleading for a reader, since it's not actually the creature that's going to be targetted for the damage. You can just use `owner` in the `atStartOfTurn` implementation.

src/main/java/stsjorbsmod/powers/PossessionPower.java
```diff
@@ -0,0 +1,63 @@
...
+        this.isTurnBased = true;
+
+        updateDescription();
+    }
+
+    @Override
+    public void atStartOfTurn() {
+        AbstractMonster m = this.target;
+        JorbsMod.logger.info("m and this.owner: " + m);
```
dbjorge, line 31: I think the logging statements can probably be removed after you're comfortable the power is working correctly; no need to spam the log with them if we aren't actively debugging

src/main/java/stsjorbsmod/powers/PossessionPower.java
```diff
@@ -0,0 +1,63 @@
...
+
+        while (m == this.target) {
+            m = AbstractDungeon.getMonsters().getRandomMonster((AbstractMonster) null, true, AbstractDungeon.cardRandomRng);
+        }
+        JorbsMod.logger.info("monster chosen: " + m);
+        addToBot(new DamageAction(m, new DamageInfo(this.owner, this.owner.maxHealth), AbstractGameAction.AttackEffect.SLASH_DIAGONAL));
+        JorbsMod.logger.info("Did damage");
+        if (this.amount == 2) {
```
dbjorge, line 39: I think it would be better to use a loop over `this.amount` rather than repeating the code in an if block like this; repeating the code is redundant, and won't work correctly if more than one `Possession+` card is stacked on the same monster

dbjorge (dismissed):
Thanks for the contribution! Some comments inline

src/main/java/stsjorbsmod/cards/cull/deckoftrials/Possession.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+        else {
+            AbstractDungeon.effectList.add(new ThoughtBubble(AbstractDungeon.player.dialogX, AbstractDungeon.player.dialogY, 3.0F, "Only one enemy!", true));
```
wang429, line 47: For localization purposes, the string "Only one enemy!" can be moved to the card string file using `EXTENDED_DESCRIPTION`

src/main/java/stsjorbsmod/powers/PossessionPower.java
```diff
@@ -0,0 +1,63 @@
...
+    public void atStartOfTurn() {
+        AbstractMonster m = this.target;
+        JorbsMod.logger.info("m and this.owner: " + m);
+
+        while (m == this.target) {
+            m = AbstractDungeon.getMonsters().getRandomMonster((AbstractMonster) null, true, AbstractDungeon.cardRandomRng);
+        }
+        JorbsMod.logger.info("monster chosen: " + m);
+        addToBot(new DamageAction(m, new DamageInfo(this.owner, this.owner.maxHealth), AbstractGameAction.AttackEffect.SLASH_DIAGONAL));
+        JorbsMod.logger.info("Did damage");
```
wang429, line 38: Note that DamageAction doesn't trigger until the power is done resolving.

wang429:
@nic3245 sorry about not noticing this before. @acourrau was wondering about this PR. He noticed that `possession` branch had changes that were made in line with our suggestions. So, we looked at your repo in GitHub and saw there are two branches for "possession" in your repository, `Possession` and `possession`. Apparently, the Git CLI does some case insensitivity and only shows the one branch. However, the branch this PR is sourcing from isn't the branch you updated. If you're still around please either update the source branch for this PR or maybe reopen one. We didn't mean to ignore you at all.

acourrau:
Closing PR in favor of the one that has your updated changes fixing comments from brosmike.  Thank you very much for the work, and sorry we missed this before!

PR #512, opened 2020-05-08
Reap and Sow

acourrau:
Added Reap and Sow. New save data to store the value between fights. New power (mostly for a visual reminder).


src/main/java/stsjorbsmod/JorbsMod.java
```diff
@@ -456,4 +457,14 @@ public void receiveStartGame() {
             }
         }
     }
+
+    @Override
+    public void receiveOnBattleStart(AbstractRoom abstractRoom) {
+        if (ReapAndSowSaveData.reapAndSowDamage != 0) {
+            AbstractDungeon.actionManager.addToBottom(
+                    new DamageAllEnemiesAction(null, DamageInfo.createDamageMatrix(ReapAndSowSaveData.reapAndSowDamage, true),
```
dbjorge, line 465: I think probably it would be better to use the `AbstractDungeon.player` as the damage source rather than null; it's still the player responsible for the damage, just on a delay

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -229,6 +229,12 @@
       "Whenever you apply #yBurning, apply that much #yBurning #b%1$s extra times."
     ]
   },
+  "stsjorbsmod:ReapAndSowPower": {
+    "NAME": "Reap and Sow",
+    "DESCRIPTIONS": [
+      "At the start of the next combat, attack ALL enemies for #b%1d damage."
```
dbjorge, line 235: Probably best to avoid the term `attack` here, it might imply to the player that it should work like attack damage and apply vulnerable/weak/etc. I think it's probably better to implement it like you did and use non-attack damage (via the THORNS damageType), and update the wording to be more like `deal #b%1d damage to ALL enemies.`

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -617,6 +617,10 @@
     "DESCRIPTION": "Deal !D! damage. NL Apply !M! Weak.",
     "UPGRADE_DESCRIPTION": "Deal !D! damage. NL Apply !M! Weak !stsjorbsmod:MetaMagic! times."
   },
+  "stsjorbsmod:ReapAndSow": {
+    "NAME": "Reap and Sow",
+    "DESCRIPTION": "Deal !D! damage, if Fatal, deal the leftover damage to ALL enemies at the start of the next combat."
```
dbjorge, line 622: The comma between "damage" and "if" feels like it's creating a sentence fragment; I think a period would probably work better there.

src/main/java/stsjorbsmod/powers/ReapAndSowPower.java
```diff
@@ -0,0 +1,29 @@
+package stsjorbsmod.powers;
+
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.powers.AbstractPower;
+import stsjorbsmod.characters.ReapAndSowSaveData;
+
+
+public class ReapAndSowPower extends CustomJorbsModPower {
```
dbjorge, line 8: I really like the use of the power as a visual indicator! However, consider that there are effects that can remove buffs from the character (particularly, the `Amnesia` curse), and that the player could theoretically stack multiple instances of `ReapAndSow`; I think I would suggest using the power rather than the savedata to track the overkill damage while the combat is still happening and handle the stacking behavior, and only commit the power amount to `ReapAndSowSaveData.reapAndSowDamage` at the end of combat (I think `onVictory` is the hook method you'd want, but you'd want to double check the ordering for that vs save data being saved).

src/main/java/stsjorbsmod/cards/cull/ReapAndSow.java
```diff
@@ -0,0 +1,61 @@
...
+    public ReapAndSow() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        baseDamage = damage = DAMAGE;
+        magicNumber = baseMagicNumber = 0;
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        overkill = getOverKill(m);
+        if (overkill > 0 && m.hasPower("Minion")) {
```
dbjorge, line 39: This is missing a few edge cases (eg, enemies with half-dead states like awakened one and darklings). I think the best option would be to use `DamageWithOnKillEffectAction` instead of a plain `DamageAction` and move the `ReapAndSowPower` application to the on-kill callback passed to that action.

src/main/java/stsjorbsmod/cards/cull/ReapAndSow.java
```diff
@@ -0,0 +1,61 @@
...
+            addToBot(new ApplyPowerAction(p, p, new ReapAndSowPower(p, overkill)));
+        }
+        addToBot(new DamageAction(m, new DamageInfo(p, damage), AbstractGameAction.AttackEffect.SLASH_DIAGONAL));
+    }
+
+    private int getOverKill(AbstractMonster mo) {
+        super.calculateCardDamage(mo);
+        if (this.damage > mo.currentHealth) {
+            this.overkill = this.damage - mo.currentHealth;
```
dbjorge, line 48: I think this is probably a little too simplified; for example, I don't think it's going to do the right thing against enemies with block or vulnerable.

I think the best approach here would probably be calculate this difference after all the damage-modifying effects are applied in `AbstractMonster.damage`; the `lastDamageTaken` field that method sets is *almost* exactly what we want, if only it didn't get clamped to the monster's current help. Since we don't want to rewrite all the calculation logic and the values we need are lost by the time `AbstractMonster.damage` is done, I'm not sure there's a better option than adding a patch here. I'd probably lean towards adding an `AbstractMonster` patch which:
* Adds a `SpireField` to `AbstractMonster` for `lastOverkillDamage`, which defaults to 0
* Adds an `Instrument` patch for `AbstractMonster.damage` that replaces the `this.lastDamageTaken` field access with `$_ = $proceed(); lastOverkillDamageField.set(__this, damageAmount - __this.lastDamageTaken);`

src/main/java/stsjorbsmod/cards/cull/ReapAndSow.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final int COST = 2;
+    private static final int DAMAGE = 25;
+    private static final int UPGRADE_DAMAGE = 8;
+
+    private int overkill = 0;
+
+    public ReapAndSow() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        baseDamage = damage = DAMAGE;
+        magicNumber = baseMagicNumber = 0;
```
dbjorge, line 33: You can leave this out if the card doesn't use magicNumber

src/main/java/stsjorbsmod/cards/cull/ReapAndSow.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final CardRarity RARITY = CardRarity.RARE;
+    private static final CardTarget TARGET = CardTarget.ENEMY;
+    private static final CardType TYPE = CardType.ATTACK;
+    public static final CardColor COLOR = Cull.Enums.CULL_CARD_COLOR;
+
+    private static final int COST = 2;
+    private static final int DAMAGE = 25;
+    private static final int UPGRADE_DAMAGE = 8;
+
+    private int overkill = 0;
```
dbjorge, line 28: Does this have to be a field? It looks like it's passed around via return values and local variables

dbjorge (dismissed):
Thanks for the PR! It makes me happy to see you progressing onto more challenging cards :)

Some suggestions inline

src/main/java/stsjorbsmod/JorbsMod.java
acourrau: makes perfect sense.

wang429 (approved):

wang429:
Looks good to me!

PR #513, opened 2020-05-09
Sacrifice

acourrau:
Added Sacrifice Card to Deck of Trials. Works well enough, the X cost card workaround is a little odd, but the "freeToPlayOnce" does not affect X cost cards.


src/main/java/stsjorbsmod/cards/cull/deckoftrials/Sacrifice.java
```diff
@@ -0,0 +1,52 @@
...
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        baseMagicNumber = magicNumber = ENERGY_GAIN;
+
+        this.isEthereal = true;
+        SelfExertField.selfExert.set(this, true);
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster abstractMonster) {
+        addToBot(new ApplyPowerAction(p, p, new SacrificePower(p, magicNumber)));
```
dbjorge, line 35: I think this should be using `1`, not `magicNumber` (but I appreciate that the power handles the >1 case, I think that's a reasonable choice for handling stacking multiple copies of Sacrifice)

src/main/java/stsjorbsmod/powers/SacrificePower.java
```diff
@@ -0,0 +1,63 @@
...
+        super.onUseCard(card, action);
+
+        int hpCost = 0;
+        if (card.costForTurn != -1) {
+            hpCost = card.costForTurn;
+            card.freeToPlayOnce = true;
+        }
+        else {
+            hpCost = AbstractDungeon.player.energy.energy;
+            addToBot(new GainEnergyAction(hpCost)); // Makes X cost cards "free"
```
dbjorge, line 36: I would have thought you'd be able to use `card.freeToPlayOnce` instead of this for X cost cards, based on looking over a handful of the base game cases; I think that would probably be better for cooperating with other freeToPlayOnce based effects

dbjorge: Ah, I see, it's an ordering issue; `AbstractPlayer.useCard` calls `AbstractCard.use` (which reads the `freeToPlayOnce` value) *before* it constructs the corresponding `UseCardAction` (which invokes our power's callback). Hmm.

acourrau: Yeah, I agree though I don't like the implementation...

src/main/java/stsjorbsmod/powers/SacrificePower.java
```diff
@@ -0,0 +1,63 @@
...
+    public SacrificePower(AbstractPlayer owner, int stacks) {
+        super(STATIC);
+        this.isTurnBased = true;
+        this.amount = stacks;
+        this.owner = owner;
+        this.updateDescription();
+    }
+
+    @Override
+    public void onUseCard(AbstractCard card, UseCardAction action) {
```
dbjorge, line 26: I think this implementation is missing a few interactions with other effects that also mess with how much cards cost; in particular, `AbstractPlayer.useCard` has a variety of other effects besides just `costForTurn` that would normally make using a card free, but which this implementation is still going to cost max HP for.

I think the two options for dealing with this would be:
* Repeat the logic from `AbstractPlayer.UseCard` L1693-1694 here to allow 0 hpCost, or
* Scrap `onUseCard` and instead implement the power's logic using an `EnergyManager.use` patch

I would actually lean towards the second option for this case, even though it does involve a new patch. The two main reasons I think the patch might be better in this case are:

* It also avoids the problem of needing to do a hacky thing to handle the X cost card case (X cost cards also use `EnergyManager.use` and already have logic for not calling it if they're supposed to be free to play, so I think the patch solution would just work naturally for them)
* It cooperates better with any other mods that might add Corruption-like effects by patching `AbstractPlayer.useCard`

dbjorge (dismissed):
Thanks for the PR! Left a design suggestion

src/main/java/stsjorbsmod/powers/SacrificePower.java
dbjorge: If we do the patch, probably best to update the wording on the description to avoid the "to play cards" part, since in theory it would also affect any other cases that cost energy outside of playing a card (I don't think there are currently any such cases in the base game or our mod, but maybe there are other mods that add them)

wang429 (approved):
Looks good. Thanks!

PR #514, opened 2020-05-10
Clarify

acourrau:
Added Clarify Card, followed Rebuttal as an example. Had to write a patch for DrawPileViewScreen, frozen eye is hard coded in there. See:
```    
if (!AbstractDungeon.player.hasRelic("Frozen Eye")) {
   this.drawPileCopy.sortAlphabetically(true);
   this.drawPileCopy.sortByRarityPlusStatusCardType(true);
}
```


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -120,6 +120,11 @@
     "NAME": "Chill Touch",
     "DESCRIPTION": "Deal !D! damage to an enemy. NL It can no longer heal."
   },
+  "stsjorbsmod:Clarify": {
+    "NAME": "Clarify",
+    "DESCRIPTION": "Retain. While this is in your hand, Retain !M! extra card. NL stsjorbsmod:Exert. NL Gain Frozen Eye Effect.",
```
dbjorge, line 125: There are enough cards with this pattern that we've started establishing a unique grammar for it diverged a bit from the base game to fit it better on the cards (see Frustration and Rebuttal). Let's do:

```Retain: While this is in your hand, Retain !M! extra card. NL stsjorbsmod:Exert: Your draw pile is shown in order.```

File: src/main/java/stsjorbsmod/patches/ClarifyPatch.java [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+    @SpirePatch(clz = DrawPileViewScreen.class, method = "open")
+    @SpirePatch(clz = DrawPileViewScreen.class, method = "render")
+    public static class AbstractRoomPatch {
+        public static ExprEditor Instrument() {
+            final int[] counter = {0};
+            return new ExprEditor() {
+                @Override
+                public void edit(MethodCall methodCall) throws CannotCompileException {
+                    if (methodCall.getMethodName().equals("hasRelic")) {
+                        if (counter[0] == 0) {
```
dbjorge, line 23: What is the counter for?

src/main/java/stsjorbsmod/patches/ClarifyPatch.java
```diff
@@ -0,0 +1,37 @@
...
+                            methodCall.replace(
+                                    "$_= $proceed($$) || "+ClarifyPatchName+".playerHasPower();"
```
dbjorge, line 25: Today, Frozen Eye is the only relic the base game is calling `hasRelic` for in these methods, but it would probably be a good idea to make the patch a little more specific and only apply specifically to `hasRelic("Frozen Eye")` calls in case a different check gets added in the future (or by a patch from a different mod).

You can do that using the same pattern that `PlayableCursesPatch` uses, something like:

```java
        "{ $_ = ($proceed($$) || (%1$s.isFrozenEye($1) && %1$s.hasClarifyPower($0))); }",
        ClarifyPatch.class.getName()));
```

src/main/java/stsjorbsmod/patches/ClarifyPatch.java
acourrau: Probably nothing. LOL whoops, leftover from testing

src/main/java/stsjorbsmod/cards/cull/Clarify.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public Clarify() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        selfRetain = true;
+        baseMagicNumber = magicNumber = CARDS_RETAINED;
+        SelfExertField.selfExert.set(this, true);
+    }
+
+    @Override
+    public void triggerWhenDrawn() {
```
dbjorge, line 38: I feel a little uneasy about the "apply RetainCardPower when drawn and remove it when leaving hand" pattern:
* I think it's confusing for Amnesia to be able to remove it
* I think there are some edge case ways in which it can leave your hand that don't have hooks (eg, stuff that moves cards to the top of your deck)

I think it might be better to use `triggerOnEndOfTurnForPlayingCard` to queue a `RetainCardsPolitelyAction`; it doesn't cooperate quite as well with multiple copies of Clarify (it'll give 2 separate dialogs), but it doesn't outright break in any cases I'm thinking of

dbjorge (dismissed):
Thanks for the PR! Suggestions inline

wang429:
hmm, can't change this one until brosmike clears his requesting changes.

PR #516, opened 2020-05-15
Old Book

acourrau:
~~Old Book added. Card is a little strange as it's supposed to only work when it's in your hand, but you discard your hand before you receive attack damage. Adds a power that lasts till end of round unless consumed.~~
See comment below on reopen of PR.


acourrau:
Misunderstood how card was supposed to work. Fixing.

acourrau:
Fixed how book was supposed to function. Now starts Entombed. When the player would die, it exhumes then destroys the card (just for effects).

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -182,6 +182,13 @@
       "You cannot remember other memories until you have #yClarity of your current memory."
     ]
   },
+  "stsjorbsmod:OldBookPower": {
+    "NAME": "Old Book",
+    "DESCRIPTIONS": [
+      "When you die, instead destroy Old Book.",
+      "When you die, instead destroy Old Book and heal #g%1d percent of your max HP."
```
dbjorge, line 189: Lizard tail and Fairy in a bottle uses "When you **would** die" for the conditional, probably best to be consistent

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -182,6 +182,13 @@
       "You cannot remember other memories until you have #yClarity of your current memory."
     ]
   },
+  "stsjorbsmod:OldBookPower": {
+    "NAME": "Old Book",
+    "DESCRIPTIONS": [
+      "When you die, instead destroy Old Book.",
```
dbjorge, line 188: Should probably treat "destroy" like a keyword (`#yDestroy`)

src/main/java/stsjorbsmod/powers/OldBookPower.java
```diff
@@ -0,0 +1,65 @@
...
+
+        this.isTurnBased = false;
+        this.card = card;
+        this.owner = owner;
+
+        updateDescription();
+    }
+
+    @Override
+    public int onAttacked(DamageInfo info, int damageAmount) {
```
dbjorge, line 29: I think `onAttacked` isn't quite late enough of a trigger for us to use here; it's going to happen before Torii, doesn't happen quite late enough for every edge case; I think the most serious interactions this would order incorrectly with are Tungsten Rod, Torii, and Spirit Shield. I think probably the way to go here would be to modify SpiritShieldPower to use `OnPlayerHpLossPowerSubscriber` and use `onPlayerHpLoss` then have `Old Book` use `onLoseHp`.

src/main/java/stsjorbsmod/powers/OldBookPower.java
```diff
@@ -0,0 +1,65 @@
...
+        if (damageAmount >= owner.currentHealth) {
+            info.output = damageAmount = 0;
```
dbjorge, line 31: I think this should be returning 0 in this path, rather than modifying `info.output`

src/main/java/stsjorbsmod/powers/OldBookPower.java
```diff
@@ -0,0 +1,65 @@
...
+public class OldBookPower extends CustomJorbsModPower {
+    public static final StaticPowerInfo STATIC = StaticPowerInfo.Load(OldBookPower.class);
+    public static final String POWER_ID = STATIC.ID;
+
+    private AbstractCard card;
+
+    public OldBookPower(final AbstractCreature owner, AbstractCard card) {
+        super(STATIC);
+
+        this.isTurnBased = false;
```
dbjorge, line 21: I like using a power to make it more obvious that the effect is in play, but I think it should probably not be dismissable by Amnesia, so it should probably use a `SPECIAL` type to avoid that.

src/main/java/stsjorbsmod/powers/OldBookPower.java
```diff
@@ -0,0 +1,65 @@
...
+import stsjorbsmod.actions.ExhumeCardsAction;
+import stsjorbsmod.util.CardMetaUtils;
+
+public class OldBookPower extends CustomJorbsModPower {
+    public static final StaticPowerInfo STATIC = StaticPowerInfo.Load(OldBookPower.class);
+    public static final String POWER_ID = STATIC.ID;
+
+    private AbstractCard card;
+
+    public OldBookPower(final AbstractCreature owner, AbstractCard card) {
```
dbjorge, line 18: I'm not sure this is going to do the right thing if there are multiple copies of `Old Book` in the player's deck; it's probably best to have one stacking power rather than multiple distinct powers, since we'd only want one book to trigger at a time.

src/main/java/stsjorbsmod/cards/cull/OldBook.java
```diff
@@ -0,0 +1,59 @@
...
+        EntombedField.entombed.set(this, true);
+
+        magicNumber = baseMagicNumber = HEAL_PERCENT;
+
+        tags.add(LEGENDARY);
+    }
+
+    @Override
+    public void onCardEntombed() {
+        addToTop(new ApplyPowerAction(p, p, new OldBookPower(AbstractDungeon.player, this)));
```
dbjorge, line 41: Should probably also have an `onCardExhumed` condition to reduce the power, to in case some other effect manages to exhume the card somehow (eg, the player somehow obtains a copy of the card Exhume).

src/main/java/stsjorbsmod/powers/OldBookPower.java
```diff
@@ -0,0 +1,65 @@
...
+                float percent = (float)card.magicNumber / 100.0F;
+                int healAmt = (int)((float)owner.maxHealth * percent);
+                if (healAmt < 1) {
+                    healAmt = 1;
+                }
+
+                owner.heal(healAmt, true);
+            }
+
+            AbstractDungeon.player.masterDeck.removeCard(card);
```
dbjorge, line 45: `destroyCardPermanently` already covers removing the masterDeck version

src/main/java/stsjorbsmod/powers/OldBookPower.java
```diff
@@ -0,0 +1,65 @@
...
+        this.owner = owner;
+
+        updateDescription();
+    }
+
+    @Override
+    public int onAttacked(DamageInfo info, int damageAmount) {
+        if (damageAmount >= owner.currentHealth) {
+            info.output = damageAmount = 0;
+            addToBot(new ExhumeCardsAction(card));
```
dbjorge, line 32: I don't think this is actually going to do anything; `destroyCardPermanently` is going to *immediately* remove the card from the exhaust pile, but this action is only going to search the exhaust pile for the card to exhume later, once the action is pulled off the queue.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -516,6 +516,11 @@
     "NAME": "Oak Leaf",
     "DESCRIPTION": "Gain !M! Dexterity."
   },
+  "stsjorbsmod:OldBook": {
+    "NAME": "Old Book",
+    "DESCRIPTION": "stsjorbsmod:Entombed (Die). NL. Unplayable. NL stsjorbsmod:Destroy this.",
```
dbjorge, line 521: I think the `Entombed (Die)` syntax was just shorthand in the spreadsheet; the parenthetical isn't something we've used in other cards, so I'm not sure this will necessarily be clear to a player. I'd suggest something like:

```stsjorbsmod:Entombed. NL stsjorbsmod:Exhume when you would die; instead, stsjorbsmod:Destroy this card. NL Unplayable.```

**Review by dbjorge — DISMISSED** [review]
Thanks! Some suggestions inline

File: src/main/java/stsjorbsmod/powers/OldBookPower.java (hunk 3 above)
Comment (dbjorge) [reply src/main/java/stsjorbsmod/powers/OldBookPower.java:29]: Jorbs just suggested that he'd prefer Old Book to trigger *after* Lizard Tail and Fairy in a Bottle; that makes it a little more awkward, since both of those don't trigger until well after damage is applied, but Old Book wants to be cancelling the damage the attack *would* apply. The options would be:

* Patch in some sort of `OnPlayerDeathToPrevent` hook in `AbstractPlayer.damage` just after Lizard Tail and Fairy in a Bottle, and have it heal the player for as much damage as they just took when triggered
* Keep it like this comment originally suggested, but have it check whether Fairy/Lizard *would* be applying and skip itself if it thinks they will apply later

I think I like the second option more, even though it requires a gross amount of hardcoding for Mark of the Bloom and the other 2 effects, because the former option creates a weird/confusing behavior in combination with powers/etc that affect healing (eg, it won't do the right thing if Mark of the Bloom is active)

File: src/main/java/stsjorbsmod/powers/OldBookPower.java (hunk 5 above)
Comment (wang429) [reply src/main/java/stsjorbsmod/powers/OldBookPower.java:24]: resolving conversation as this has been addressed.

File: src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json [hunk 11]
```diff
@@ -189,6 +189,13 @@
       "You cannot remember other memories until you have #yClarity of your current memory."
     ]
   },
+  "stsjorbsmod:OldBookPower": {
+    "NAME": "Old Book",
+    "DESCRIPTIONS": [
+      "When you would die, instead #yDestroy Old Book.",
+      "When you would die, instead #yDestroy Old Book+ and heal #g%1d percent of your max HP."
```
wang429, line 196: It might be a good idea for card references to have a color as well. Not sure if we did anything with Wrath Memory.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -537,6 +537,11 @@
     "NAME": "Oak Leaf",
     "DESCRIPTION": "Gain !M! Dexterity."
   },
+  "stsjorbsmod:OldBook": {
+    "NAME": "Old Book",
+    "DESCRIPTION": "stsjorbsmod:Entombed. NL stsjorbsmod:Exhume when you would die; instead stsjorbsmod:Destroy this card. NL Unplayable",
+    "UPGRADE_DESCRIPTION": "stsjorbsmod:Entombed. NL stsjorbsmod:Exhume when you would die; instead stsjorbsmod:Destroy this card and heal !M! percent of your Max HP. NL Unplayable."
```
wang429, line 543: Lizard Tail and Fairy in a Bottle both are `#bNum%`

wang429:
Oh, could you update CHANGELOG.md to include this card?

wang429 (approved):
Looks good. Thanks!

PR #520, opened 2020-05-23
Coup de grace

acourrau:
Oh boy. Big card.

So there are a few unique things about this. First of all if you (or the enemy) have the CoupDeGracePower, damage values will be displayed WITHOUT intangible reduction. E.G. 9 incoming damage with 1 intangible still shows 9 incoming damage instead of 1. Same is true for withering debuff. 9 incoming damage with 1 intangible and 2 withering shows 9 damage instead of 3. I don't love the way this displays, I'd love to see if we could add a secondary display value that would show the "Real" damage you're going to take in addition to the total incoming damage. 

Those are the big parts of the patching. WitheringPatch and CoupDeGrace patch are both slightly affected. 

The other weird thing about this card is that it can target the player OR the enemy. It checks what's being hovered and sets the Target flag appropriately. Ideally I'd like to have an arrow point to the player when you're not hovering it over an enemy. 

Perhaps both those changes can come in a later version once this has been playtested and balanced, or if there's a simple solution available then we can add it quickly.

Thanks!


src/main/java/stsjorbsmod/patches/CoupDeGracePatch.java
```diff
@@ -0,0 +1,61 @@
...
+import com.evacipated.cardcrawl.modthespire.lib.SpirePrefixPatch;
+import com.evacipated.cardcrawl.modthespire.lib.SpireReturn;
+import com.megacrit.cardcrawl.cards.DamageInfo;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import com.megacrit.cardcrawl.powers.IntangiblePlayerPower;
+import stsjorbsmod.powers.CoupDeGracePower;
+
+public class CoupDeGracePatch {
```
wang429, line 13: For the monster intent, you might be able to get away with patching `AbstractMonster::calculateDamage` such that you wrap the existing `this.intentDmg = dmg;` with something like 
```
if shouldCalculatePreentedDamage
  intentDmg = 1
else
  proceed
```

src/main/java/stsjorbsmod/cards/cull/CoupDeGrace.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        if (AbstractDungeon.player.hasRelic(ChemicalX.ID)) {
+            AbstractDungeon.player.getRelic(ChemicalX.ID).flash();
+        }
+
+        if (targetingEnemy && m != null) {
+            addToBot(new ApplyPowerAction(m, m, new IntangiblePlayerPower(m, magicNumber)));
```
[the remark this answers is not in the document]
acourrau: IntangiblePower works slightly differently IIRC, I think it's hardcoded to apply every other turn. We could certainly use that, I'd just have to change the code. IntangiblePlayerPower is misleading with it's name for sure.

src/main/java/stsjorbsmod/cards/cull/CoupDeGrace.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public CoupDeGrace() {
+        super(ID, COST, TYPE, COLOR, RARITY, TARGET);
+        magicNumber = baseMagicNumber = BASE_EXTRA_PLAYS;
+        tags.add(LEGENDARY);
+    }
+
+    @Override
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        if (AbstractDungeon.player.hasRelic(ChemicalX.ID)) {
```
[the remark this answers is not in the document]
acourrau: Sounds good, I'll look at those and adjust accordingly.

src/main/java/stsjorbsmod/patches/CoupDeGracePatch.java
```diff
@@ -0,0 +1,61 @@
...
+        @SpirePrefixPatch
+        public static SpireReturn patch(IntangiblePlayerPower __this, float damage, DamageInfo.DamageType type) {
+            if (!shouldCalculatePreventedDamage(__this.owner)) {
+                SpireReturn.Continue();
+            }
+            return SpireReturn.Return(damage);
+        }
+    }
+
+    public static boolean shouldCalculatePreventedDamage(AbstractCreature m) {
```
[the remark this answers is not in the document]
acourrau: Agreed.

src/main/java/stsjorbsmod/patches/CoupDeGracePatch.java
acourrau: Ohh, that's a path I hadn't considered. I'll look into this.

acourrau:
Thanks for comments!

wang429:
There are still comment on this, please address them, either with comments on won't do or pushing changes. Thanks.

src/main/java/stsjorbsmod/cards/cull/CoupDeGrace.java
```diff
@@ -0,0 +1,101 @@
...
+//        if (AbstractDungeon.player.hasRelic(ChemicalX.ID)) {
+//            effect += 2;
+//        }
+//
+//        if (this.upgraded) {
+//            effect += 1;
+//        }
+//
+//        return effect;
+//    }
```
wang429, line 92: Please remove commented code.

src/main/java/stsjorbsmod/patches/CoupDeGracePatch.java
wang429: I'm curious if you thought more about this. _Not that I quite remember what I was thinking_ lol

wang429 (changes requested):
Some minor comments in line. Thanks!

src/main/java/stsjorbsmod/cards/cull/CoupDeGrace.java
acourrau: Ty, forgot about this.

wang429 (approved):
I think it looks alright. Thanks!

PR #521, opened 2020-07-02
5 new card arts

neablis-7:
Frantic Search, Wanderer Bad uncommon, final art from Chevy.
Refuse to Forget, Wanderer manipulation uncommon, final art from Chevy. 
Animate Objects, Wanderer AOE uncommon, final art from Chevy. 
Index, Wanderer manipulation common, art is from lobbien on the discord. 
Coda, Cull card, art is from Christina Oran via google photos. (I did confirm that the full name was how they wished to be credited)


wang429:
Note to self, update changelog with card art credits.

PR #526, opened 2020-08-13
Fix tooltips and about.md to have the correct value of 10 for Greed

modargo:
I noticed this while playing. Greed was changed to only give 10 gold a while back, but the card keyword tooltip wasn't updated for most languages, and also hadn't been updated on the "about" page. See screenshot, which shows how Locate Object has an inaccurate tooltip.

The value was already correct for French, so that's the only language file that I didn't change.

image


acourrau:
Thanks for the fix. Merging.

modargo:
Great -- thank you.

PR #532, opened 2020-10-05
1 new card art

neablis-7:
Art for Inferno, from Chevy.


acourrau (approved):
Thanks for the great artwork and thanks for helping merge!

PR #534, opened 2020-10-13
Update zhs translation: new cards and wording changes

yanang007:
Translation for new cards:
 CoupDeGrace, FocusedRage,

Translation for new powers:
 CoupDeGracePower, OldBookPower, PossessionPower


acourrau:
Just a heads up, I made some slight changes to the highlighting and wording of old book. I imagine the translations will need a cursory look. I'll hold off making any formal requests for updates until coup de grace gets merged, as changes are a possibility.

Thanks for the work! Very much appreciated!

yanang007:
Got it! I will keep track on those new changes!
But I accidentally broke my repo... So I made another fork which left this PR invalid now. I probably need to close it and open another one when appropriate. Sorry for that...

acourrau:
Not a problem at all! We're close on the merge for Coup De Grace I think, so it shouldn't be too long. We did add some new keywords for Coup De Grace, and it also affects an old card (True Strike).

acourrau:
@yanang007 we're looking really good as far as progress goes. I think it's probably safe to update these translations.

PR #538, opened 2020-10-27
Possession

acourrau:


acourrau:
See comments from https://github.com/dbjorge/jorbs-spire-mod/pull/511 - mixup with local branch names, this one has all the comments that were requested by brosmike in first one.

wang429 (changes requested):
I left some inline comments. Looks good otherwise

A couple non-review things:
1. Please address the conflict against master.
2. It would be nice to have a special "Possessed" intent with the damage value there. Do we have an issue for that?

PR #540, opened 2020-10-27
Warped Glass

wang429:
From the spreadsheet:
Relic
Uncommon
Deal 3 damage to all enemies when you exert a card.
> this could be same relic as mindglass but older

Probably art direction.


acourrau (approved):
Looks great! Everything seems to work fine. Great addition.

PR #542, opened 2020-10-27
Possession

acourrau:
This contains all comments from PR #511 and #538


src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Power-Strings.json
```diff
@@ -245,6 +245,13 @@
       "The next card you play this turn is played an additional #b%1$s times."
     ]
   },
+  "stsjorbsmod:PossessionPower": {
+    "NAME": "Possessed",
+    "DESCRIPTIONS": [
+      "This turn instead of attacking, this creature will attack a random ally for #b%1$s damage.",
```
wang429, line 251: Thinking about it a bit more, enemy intents aren't always just attacking. Perhaps "This turn, instead of taking its normal turn,..."? I dunno, that feels clunky. Ideally, we could check the intent and swap "attacking" with "buffing", "debuffing", or whatever is good text for "???". These texts would also be localized.

I would prefer these singular and plural be consistent.

wang429 (changes requested):
Looking really good! Just noticed that I don't see the changelog here. Could you update that too?

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Keyword-Strings.json
```diff
@@ -118,6 +118,10 @@
     "NAMES": ["patience"],
     "DESCRIPTION": "Gain 1 #yCoil each time you play a card. When you remember a new memory, deal 1 damage to ALL enemies for each #yCoil and lose all #yCoil."
   },
+  {
+    "NAMES": ["possessed"],
+    "DESCRIPTION": "#yPossessed creatures will deal damage equal to it's max HP to a random ally for each #yPossessed stack."
```
wang429, line 123: "its"

wang429 (changes requested):
Just the one trivial comment. Thanks for pushing through these reviews!

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Keyword-Strings.json
acourrau: fixed. Also adjusted "max HP" to be consistent across all of our cards

wang429 (approved):
Looks good!

PR #499, opened 2020-04-29
ZHS translation update: CULL Voiceovers and other changes

yanang007:
Translation for CULL Voiceovers and new things.
Keyword renaming: Manifest
Layout changes and conformance improvements.

PR #547, opened 2020-10-29
Update zhs translation

yanang007:
ZHS translation for upcoming cards, keywords, powers, relics and UI contents.

And an overall check is made to keep everything up to date.

### DIFF
--- a/src/main/resources/stsjorbsmodResources/localization/zhs/JorbsMod-Card-Strings.json
+++ b/src/main/resources/stsjorbsmodResources/localization/zhs/JorbsMod-Card-Strings.json
@@ -134,8 +134,8 @@
   },
   "stsjorbsmod:Clarify": {
     "NAME": "澄澈",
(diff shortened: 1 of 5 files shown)


acourrau (approved):

acourrau:
Thank you for grabbing all those changes!

acourrau:
Would you be able to add ZHS for the `JorbsMod-Cull-Voiceover-Strings.json` file?

yanang007:
@acourrau 

Yes, but it seems that the translation has already been update in #499 . I'm wondering is there any update that I missed?
Thank you for the heads-up.

acourrau:
Forgive me, they are there I missed that. We did add 2 new things, shadow relic and repressed memory card (including a new power for the card). Would you look at those?

yanang007:
Not a problem at all!
I have included translation for some of upcoming changes in the earlier update while they were still in PR at that point. However, I'm not at home right now, so it may take minutes to keep up with new changes.

acourrau:
You're a machine, thank you!

PR #556, opened 2020-11-02
Sacrifice fix

### LINKED ISSUE #553: Cards cannot be played via Sacrifice with 0 energy.
**Version**
Slay the spire: MAIN
Mod version: 1.9.0 

**Describe the bug**
When you have 0 energy, you are not able to play cards via Sacrifice.

**To Reproduce**
1. Enter combat with Sacrifice in hand.
2. Play cards to use all your energy (not including Sacrifice)
3. Use Sacrifice
4. No cards are playable

**Expected behavior**
Sacrifice is meant to cost hp instead of energy to play cards, even with 0 energy

Thanks to Coogrr for reporting.

acourrau:
Fixes #553 - uses BaseMod interface for replacing energy with some alternate cost. See: https://github.com/daviscook477/BaseMod/wiki/CardModifiers

Removed now deprecated SacrificeEnergyPatch, as the interface covers the need for that.


src/main/java/stsjorbsmod/powers/SacrificePower.java
```diff
@@ -42,4 +44,22 @@ public void updateDescription() {
...
+    @Override
+    public int getAlternateResource(AbstractCard c) {
+        if (c.cost == -1)
+            return EnergyPanel.totalCount;
+        return owner.maxHealth;
+    }
+
+    @Override
+    public int spendAlternateCost(AbstractCard c, int i) {
+        addToBot(new DecreaseMaxHpAction(owner, owner, i, AbstractGameAction.AttackEffect.POISON));
```
wang429, line 57: I'm not sure how much I like the losing Max Hp as a part of an action. Does the player lose Max HP if the combat ends?

src/main/java/stsjorbsmod/cards/cull/deckoftrials/Sacrifice.java
```diff
@@ -1,8 +1,11 @@
 package stsjorbsmod.cards.cull.deckoftrials;
 
+import basemod.interfaces.AlternateCardCostModifier;
 import com.megacrit.cardcrawl.actions.common.ApplyPowerAction;
 import com.megacrit.cardcrawl.actions.common.GainEnergyAction;
+import com.megacrit.cardcrawl.cards.AbstractCard;
 import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
```
wang429, line 8: could you clean these imports up?

wang429 (changes requested):
I left a couple inline comments. Also, could you update the changelog with the bugfix?

Thanks again!

src/main/java/stsjorbsmod/cards/cull/deckoftrials/Sacrifice.java
acourrau: done.

src/main/java/stsjorbsmod/powers/SacrificePower.java
acourrau: I think this is how we had it before as well as the other cards that affect Max HP, see https://github.com/dbjorge/jorbs-spire-mod/blob/master/src/main/java/stsjorbsmod/patches/SacrificeEnergyPatch.java

Play does lose the Max HP if their attack deals a killing blow, which I think is the intended effect. The player would still lose energy if that attack dealt a killing blow.

Obviously open to changes, but I was just doing a direct port over using the BaseMod features.

wang429: As long as the player loses Max HP from the card play, then I'm good.

wang429 (approved):
Looks good to me. Thanks!

PR #557, opened 2020-11-02
Adding the remove legendary cards patch to starter deck creation

### LINKED ISSUE #555: Legendary starting cards are not removed from pools until act 2 or a save and load
**Version**
Slay the spire: MAIN
Mod version: 1.9.1

**Describe the bug**
Withering, a Legendary starting card, can appear in the gremlin card matching game, allowing you to have two of them.

**To Reproduce**
1. Enter the Goblin Matching Game event
2. Flip cards until you see Withering or retry

**Expected behavior**
Legendary cards the player has a copy of should not be available in the matching game

Coogrr:
Fixes #555 
Adds a PostFix Patch to AbstractPlayer.initializeStarterDeck(), this covers removing Legendary cards in the starter deck from the pool, otherwise they are not removed by the initializeCardPools patch until act 2 or until the player save and reenters the run. They are also not removed by the Soul.Obtain patch since initializeStarterDeck does not use that method
Also changed the comment describing when AbstractDungeon.initializeCardPools() from moving between floors to between acts, which I think is correct.


wang429 (changes requested):
The changes look reasonable. Thanks for the contribution!

Please update `CHANGELOG.md` with a brief description of the fix and `docs/credits.md` with your preferred name :)

acourrau (approved):
Looks good. Thanks!

wang429 (approved):
Looks good to me! Though, I do wonder if credits should be in alphanumeric order 🤔 .

What could we ever do...

Oh, there are merge conflicts

acourrau:
I'll alphabetize the credits in a later merge.

PR #563, opened 2020-11-02
Room count increment fix

### LINKED ISSUE #560: Floor count improperly increments by 1+Manifest even when not all floors are actually moved
**Version**
Slay the spire: Main
Mod version: 1.9.1

**Describe the bug**
When Cull has manifest and moves to a floor where he's stuck, like the chest floor, the floor count still increases like he had moved the full distance

**To Reproduce**
1. Be one floor below the chest floor (8)
2. play a fight and get 2 manifest
3. Move to the chest floor (9)
4. The floor count will say 11

**Expected behavior**
The floor count should still say 9, but Manifest should go to 0 even if not all floors are moved

Coogrr:
fixes #560 
Instead of incrementing floor number by manifest directly use manifest to find next floor and use difference between next and current floor to increment floor number.
Not sure what our policy on logging is and if this is too much logging to add into one method


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Bug fixes
+
+* Fix incorrect floor number increment for Cull on chest/boss fire/boss floors
```
wang429, line 7: You're more than welcome to append your own preferred name here :)

wang429 (changes requested):
Thank you for the contribution! Could you also add yourself to `docs/credits.md`?

CHANGELOG.md
Coogrr: I added myself to credits.md on https://github.com/dbjorge/jorbs-spire-mod/pull/557, it's just not merged yet.

Coogrr: I'll add my name to the changelog as well

wang429 (approved):
Thanks! Seems alright to me.

PR #567, opened 2020-11-03
Onexertsubscriber

### LINKED ISSUE #543: Add an OnExertSubscriber
With several actions of cards/relics/etc taking effect when a card is exerted, it's probably a cleaner solution to add an OnExertSubscriber. This will future proof us and make it far easier to add these interactions in the future.

As part of this, we should make sure that the existing relics Shadow and Warped Glass are refactored to use this.

wang429:
fixes #543

No change to CHANGELOG.md because players wouldn't be generally interested.


acourrau (approved):
Seems good. Relics function as intended.

PR #573, opened 2020-11-03
Fixing Corner cases with floorCount

Coogrr:
- The chest rooms after the boss are the same y so the formula evaluates to -1 preventing increment without this if
- Every act except act 3 puts you at y=-1 before the first room, but act 3 is y=15 resulting in a reset of floor count without the if
- The Ending has an issue where boss rooms are always y=15 so it goes from y=2 to y=15 unless you stop the formula from running


wang429 (approved):
Looks good to me! Thanks for the contribution.

PR #576, opened 2020-11-09
making coup de grace rare

acourrau:


wang429 (approved):
Looks good to me. Thanks for catching this!

Issue #571, opened 2020-11-03
Spirit Shield rebalance


acourrau:
Cost: 0
Reduce all damage taken by 1 until next turn. Exhaust.

Upgrade:
Remove Exhaust.

acourrau:
fixed in commit ff86dfe

PR #578, opened 2020-11-09
rebalancing CULLs spirit shield, addresses #571

acourrau:
rebalancing CULLs spirit shield, addresses #571


wang429 (changes requested):
Thanks for taking up some of the balance changes! The code changes look good to me. Could you also update CHANGELOG.md?

wang429 (approved):
Thanks!

PR #580, opened 2020-11-11
Book of trials procs before Neow - Investment only available act 1

### LINKED ISSUE #566: Investment should only be available from deck of trials in Act 1
From Jorbs' latest feedback list:
Investment should only show up from deck of trials in act 1

### LINKED ISSUE #569: Book of Trials should proc before Neow bonus
Cards added to your deck via Book of Trials relic should be added before you choose your starting bonus from Neow.

acourrau:
Fixes #566 and Fixes #569


src/main/java/stsjorbsmod/relics/BookOfTrialsRelic.java
```diff
@@ -12,14 +12,15 @@
...
+import stsjorbsmod.cards.cull.deckoftrials.Investment;
 
 import java.util.ArrayList;
 
 import static stsjorbsmod.characters.Cull.Enums.CULL_CARD_COLOR;
 
 public class BookOfTrialsRelic extends CustomJorbsModRelic implements AtStartOfActRelicSubscriber, HasRelicStats {
     public static final String ID = JorbsMod.makeID(BookOfTrialsRelic.class);
-    private static ArrayList<String> statsCards;  // Relic Stats needs these to be static
+    public static ArrayList<String> statsCards;  // Relic Stats needs these to be static
```
wang429, line 23: I'm not seeing why this is now public.

src/main/java/stsjorbsmod/patches/NeowEventPatch.java
```diff
@@ -22,14 +36,38 @@ public static void Prefix(NeowEvent __this, boolean isDone) {
...
+    @SpirePatch(clz = NeowEvent.class, method = "buttonEffect")
+    public static class NeowEvent_Cull_AddCards {
+        public static ExprEditor Instrument() {
+            return new ExprEditor() {
+                public void edit(MethodCall methodCall) throws CannotCompileException {
+                    if (methodCall.getClassName().equals(NeowEvent.class.getName()) &&
+                            (methodCall.getMethodName().equals("blessing") || methodCall.getMethodName().equals("miniBlessing"))) {
+                        methodCall.replace(String.format(
+                                "{ %1$s.addDeckOfTrialsCardsIfNecessary(); $_ = $proceed($$); }",
+                                NeowEventPatchClass));
```
wang429, line 59: I'm not totally understanding this. How does this differ from the above patch? Should this be replacing it?

Also, if we're patching the same method, we can simply use a single static method. In this case, since they're both instrumentation patches, we can even get away with the same ExprEditor object. Part of the reason I want to patch by location rather than functionality. Because recompilation becomes quicker for the user. It does make it a bit more difficult for the devs though.
```java
                public void edit(FieldAccess fa) throws CannotCompileException {
                    if (fa.getClassName().equals(NeowEvent.class.getName()) && fa.getFieldName().equals("waitingToSave")) {
                        fa.replace(String.format(
                                "{ %1$s.doAtStartOfAct(); $_ = $proceed($$); }",
                                AtStartOfActRelicSubscriber.class.getName()));
                    }
                    if (methodCall.getClassName().equals(NeowEvent.class.getName()) &&
                            (methodCall.getMethodName().equals("blessing") || methodCall.getMethodName().equals("miniBlessing"))) {
                                "{ %1$s.addDeckOfTrialsCardsIfNecessary(); $_ = $proceed($$); }",
                                NeowEventPatchClass));
                    }
                }
```

wang429 (changes requested):
Thanks for tackling this! I left some inline comments. I do wonder if it would be better to simply remove NeowEventPatch and simply update CardCrawlGamePatch to be more general.

src/main/java/stsjorbsmod/relics/BookOfTrialsRelic.java
acourrau: Because I'm a dum dum

wang429 (approved):
Thanks for updating! Just the one comment in changelog to append yourself to the things you changed, but that's up to you to update.

PR #581, opened 2020-11-11
Siphon max hp change

### LINKED ISSUE #565: Siphon change to also grant +1 Max HP
From Jorbs' feedback list:

siphon gains 1 max hp on killing enemy

Coogrr:
Fixes #565 
Added Siphon Max HP change and infrastructure for it, Added increaseMaxHpByFlatAmountActionAndHealAction.java  to orchestrate the two action since two DamageWithOnKillEffectAction for one kill did not seem to work.


acourrau (approved):
Looks good. Minor formatting issue in 1 file.

src/main/java/stsjorbsmod/cards/cull/Siphon.java
```diff
@@ -22,19 +22,22 @@
...
         magicNumber = baseMagicNumber = HEAL_ON_FATAL;
         tags.add(CardTags.HEALING);
     }
 
     @Override
     public void use(AbstractPlayer p, AbstractMonster m) {
-        Runnable onKillEffect = () -> addToBot(new HealAction(p, p, magicNumber));
-        this.addToBot(new DamageWithOnKillEffectAction(m, new DamageInfo(p, this.damage, this.damageTypeForTurn), onKillEffect, true));
+        Runnable onKillEffect = () -> addToBot(
+                new IncreaseMaxHpByFlatAmountActionAndHealAction(p, p, magicNumber, metaMagicNumber, true));
```
wang429, line 38: Why did we choose to create an action instead of 
```java
() -> {
  addToBot(maxHpAction);
  addToBot(healAction);
}
```
brain might be farting a bit.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -741,7 +741,7 @@
   },
   "stsjorbsmod:Siphon": {
     "NAME": "Siphon",
-    "DESCRIPTION": "Deal !D! damage. NL If this kills an enemy, heal !M! HP."
+    "DESCRIPTION": "Deal !D! damage. NL If this kills an enemy, heal !M! HP and gain !stsjorbsmod:MetaMagic! Max HP."
```
wang429, line 744: I'm trying to imagine if there's any functional difference between heal then gain max hp and gain max hp then heal. I don't see any.

wang429 (changes requested):
Thanks for the contribution! I have the one comment where I'd like clarification before merging and then the one comment musing if the order matters at all.

src/main/java/stsjorbsmod/cards/cull/Siphon.java
Coogrr: I think the answer is mostly because I didn't think of it, let me try it out and see if it works as expected.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
acourrau: I considered that as well, I think they're probably the same as well.

src/main/java/stsjorbsmod/cards/cull/Siphon.java
Coogrr: Yep, works and is definitely a better solution, thanks!

wang429 (approved):
Looks good to me! Thanks for the contribution

wang429:
oops.

PR #584, opened 2020-11-16
Fixed Reap and Sow

acourrau:
updated reap and sow to work better when it ends the fight. 

Removed the power in favor of 2 actions:
 - `ReapAction` which counts up the overkill damage, using the `ActionShouldPersistPostCombatField`
 - `SowAction` which resets the values to 0 when applied. 
 -- This also uses the `ActionShouldPersistPostCombatField` in case the Sow damage ends the fight.


src/main/java/stsjorbsmod/JorbsMod.java
```diff
@@ -471,11 +472,16 @@ public void receiveCardUsed(AbstractCard c) {
 
     @Override
     public void receiveOnBattleStart(AbstractRoom abstractRoom) {
-        if (ReapAndSowSaveData.reapAndSowDamage != 0) {
-            AbstractDungeon.actionManager.addToBottom(
-                    new DamageAllEnemiesAction(AbstractDungeon.player, DamageInfo.createDamageMatrix(ReapAndSowSaveData.reapAndSowDamage, true),
-                            DamageInfo.DamageType.THORNS, AbstractGameAction.AttackEffect.SLASH_DIAGONAL));
-            ReapAndSowSaveData.reapAndSowDamage = 0;
```
wang429, line 478: We should be able to delete `ReapAndSowSaveData` as well.

src/main/java/stsjorbsmod/actions/SowAction.java
```diff
@@ -0,0 +1,25 @@
...
+import com.megacrit.cardcrawl.actions.AbstractGameAction;
+import com.megacrit.cardcrawl.actions.common.DamageAllEnemiesAction;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.cards.DamageInfo;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import stsjorbsmod.patches.ActionShouldPersistPostCombatField;
+
+public class SowAction extends AbstractGameAction {
+    public SowAction(int sowDamage) {
+        addToBot(new DamageAllEnemiesAction(AbstractDungeon.player, DamageInfo.createDamageMatrix(sowDamage, true),
```
wang429, line 12: Why is this in the constructor instead of `update()`?

src/main/java/stsjorbsmod/actions/SowAction.java
```diff
@@ -0,0 +1,25 @@
...
+                DamageInfo.DamageType.THORNS, AbstractGameAction.AttackEffect.SLASH_DIAGONAL));
+        ActionShouldPersistPostCombatField.shouldPersistPostCombat.set(this, true);
+    }
+
+    public void update() {
```
wang429, line 17: One thing I'd like to see is the ReapAndSow card flash on screen to indicate why damage occurred.

wang429 (changes requested):
Thanks for the rework! I left a few comments inline.

Don't forget to update CHANGELOG.md indicating the bugfix.

src/main/java/stsjorbsmod/JorbsMod.java
acourrau: Good catch.

src/main/java/stsjorbsmod/actions/SowAction.java
acourrau: Good question, moving.

src/main/java/stsjorbsmod/actions/SowAction.java
acourrau: Great idea!

src/main/java/stsjorbsmod/actions/SowAction.java
```diff
@@ -0,0 +1,38 @@
...
+public class SowAction extends AbstractGameAction {
+    private int sowDamage;
+
+    public SowAction(int sowDamage) {
+        this.sowDamage = sowDamage;
+        ActionShouldPersistPostCombatField.shouldPersistPostCombat.set(this, true);
+    }
+
+    public void update() {
+        AbstractCard card = new ReapAndSow();
```
wang429, line 22: It might have been better to pass along or get a copy of an instance of ReapAndSow from the deck, but this is fine.

wang429 (approved):
Looks good! Thanks for fixing this card. I did leave an inline comment, but I don't think it's particularly important.

PR #585, opened 2020-11-17
Demonic coup change

Coogrr:
Changed Demonic Coup to add 1 Ritual when retained, The Ritual doesn't activate until the following turn, which I assume is the correct behaviour. Following @wang429's suggestion on Discord for card wording. Requires Spanish localization @acourrau.


src/main/java/stsjorbsmod/cards/cull/deckoftrials/DemonicCoup.java
```diff
@@ -35,6 +37,11 @@ public void use(AbstractPlayer p, AbstractMonster m) {
         addToBot(new ApplyPowerAction(p, p, new RitualPower(p, RITUAL_AMOUNT, true)));
```
acourrau, line 37: I think you should be setting this to `magicNumber` instead of `RITUAL_AMOUNT`. I think this will always apply 2 ritual, even if it's upgraded.

acourrau (changes requested):
Great work, thanks for tackling this. One comment left inline.

src/main/java/stsjorbsmod/cards/cull/deckoftrials/DemonicCoup.java
Coogrr: I didn't even change that, but it should absolutely be magicNumber, I'll make that change

acourrau (approved):
Thanks for the quick turnaround on changes. Added localization for spanish. Approving, and leaving open for @wang429 to take a look if he'd like.

src/main/resources/stsjorbsmodResources/localization/eng/JorbsMod-Card-Strings.json
```diff
@@ -197,7 +197,7 @@
   },
   "stsjorbsmod:DemonicCoup": {
     "NAME": "Demonic Coup",
-    "DESCRIPTION": "stsjorbsmod:Ephemeral. NL Gain !M! Ritual. NL stsjorbsmod:Exert."
+    "DESCRIPTION": "Retain. NL When this card is Retained, gain !stsjorbsmod:MetaMagic! Ritual. NL Gain !M! Ritual. NL stsjorbsmod:Exert."
```
wang429, line 200: Now that I've compared against Windmill Strike, consider swapping `When this card is Retained, gain !stsjorbsmod:MetaMagic! Ritual.` and `Gain !M! Ritual.` to be consistent.

Coogrr: Like it is now?

wang429: Yeah. Though, now the Spanish should change to match.

src/main/resources/stsjorbsmodResources/localization/spa/JorbsMod-Card-Strings.json
```diff
@@ -197,7 +197,7 @@
   },
   "stsjorbsmod:DemonicCoup": {
     "NAME": "Golpe Diabólica",
-    "DESCRIPTION": "stsjorbsmod:Efímera. NL Gana !M! Ritual. NL stsjorbsmod:Expirar."
+    "DESCRIPTION": "Retiene. NL Al Retener esta carta, Gana !stsjorbsmod:MetaMagic! Ritual. NL Gana !M! Ritual stsjorbsmod:Expirar."
```
wang429, line 200: @acourrau Noticing this translation is missing an `NL` before `stsjorbsmod:Expirar`. Also, we've swapped the order of `when retained...` and `gain ritual` texts.

acourrau: Will fix.

PR #586, opened 2020-11-25
Shrieking hat

wang429:
Shrieking Hat - Unplayable, Retain, When you die instead deal all damage taken while this was in your hand to all enemies and destroy this. 
Upgrades to +3 damage

Shrieking Hat card art was already in the code base apparently.


src/main/java/stsjorbsmod/cards/cull/OldBook.java
```diff
@@ -38,13 +42,34 @@ public OldBook() {
...
+        super.triggerWhenCopied();
+        priority = currentPriority.incrementAndGet();
+    }
+
+    /**
+     * used in FastCardObtainEffect::update (combat reward cards) and ShowCardAndObtainEffect::update (just about every other card acquisition method events/transforms/etc).
+     * overloading the meaning of shrink() because lazy.
+     */
+    @Override
+    public void shrink() {
```
acourrau, line 59: I don't think this is the method you want to override as this gets called anytime a card moves from the hand to the discard pile during combat. 

In fact, I think the solution is simpler: The `masterDeck` CardGroup object holds  all cards obtained in an `ArrayList`, and the order is unaffected anywhere in the base game. We should be able to check the master deck for any `DeathPrevention` cards, and the order of the cards in that list should work great for the order in which they're obtained. Checking cards added by potions/nilrys/dead branch etc could then be checked if none are found in the `masterDeck` using your solution here. 

I think this might make the `DeathPreventionPatch` below simpler as well, as an added benefit.

src/main/java/stsjorbsmod/cards/cull/ShriekingHat.java
```diff
@@ -0,0 +1,134 @@
...
+        super.triggerWhenCopied();
+        priority = currentPriority.incrementAndGet();
+    }
+
+    /**
+     * used in ShowCardAndObtainEffect::update. Appears to be the way cards are added to the deck.
+     * overloading the meaning of shrink() because lazy.
+     */
+    @Override
+    public void shrink() {
```
acourrau, line 73: See comment above for OldBook

src/main/java/stsjorbsmod/cards/cull/ShriekingHat.java
```diff
@@ -0,0 +1,134 @@
...
+import stsjorbsmod.cards.DeathPreventionCard;
+import stsjorbsmod.cards.WasHPLostCardSubscriber;
+import stsjorbsmod.characters.Cull;
+import stsjorbsmod.util.CardMetaUtils;
+import stsjorbsmod.util.EffectUtils;
+
+import static stsjorbsmod.JorbsMod.JorbsCardTags.LEGENDARY;
+
+/**
+ * Unplayable. NL Retain. NL When you die instead deal (3 +) all damage taken while this was in your hand to all enemies. NL stsjorbsmod:Destroy.
```
acourrau, line 29: I appreciate these comments. I'd like to make this more common.

src/main/java/stsjorbsmod/patches/DeathPreventionPatch.java
```diff
@@ -13,28 +12,63 @@
...
             return false;
         if (p.hasPotion(FairyPotion.POTION_ID))
             return false;
-        if (p.hasRelic(LizardTail.ID) && ((LizardTail) p.getRelic(LizardTail.ID)).counter == -1)
+        if (p.hasRelic(LizardTail.ID) && p.getRelic(LizardTail.ID).counter == -1)
             return false;
 
-        OldBookPower po = null;
+        // requested that death preventing cards trigger in order of pickup.
+        SortedSet<DeathPreventionCard> deathPreventionCards = new TreeSet<>(Comparator.comparingInt(DeathPreventionCard::getPriority));
```
acourrau, line 40: With the change suggested above, we should be able to loop the `masterDeck` object first to determine "permanent" cards that were added to the deck that need to be consumed, and then pursue the "temporary" cards using your priority implemented.

acourrau (changes requested):
Complex card, thank you for tackling this! A few comments inline.

src/main/java/stsjorbsmod/cards/cull/OldBook.java
wang429: Good catch on shrink() being called on discard! I wonder if it's part of the animation of the card flying to a pile (draw, discard, deck)...

I'm not sure that checking the master deck is necessarily simpler. Shrieking Hat needs to be in hand in order to prevent death not just existing somewhere.

wang429: Oh i'm dumb, we already know that. Though I still don't think it's simpler.

src/main/java/stsjorbsmod/patches/DeathPreventionPatch.java
wang429: It added a little bit more complexity to the the logic, but it should work as expected.

src/main/java/stsjorbsmod/cards/cull/ShriekingHat.java
wang429: I remember doing it for some cards before, must have started to slip my mind after the break. lol

PR #588, opened 2020-12-04
zhs localizations updated

yanang007:
Sorry I'm late...


acourrau:
Not at all! We weren't waiting on you, we had just pushed out a quick crash fix due to the base game updating and breaking the mod.

acourrau (approved):
Thank you!

